PR #110, opened 2020-02-21
feat: add semantic-pull-request config

vinayakkulkarni:
- [x] Enables support for [Semantic Pull Requests](https://github.com/apps/semantic-pull-requests)


posva:
This is weird: I left a comment and closed the PR but it's not here anymore...
Thank you but I don't think we need this because we squash PRs and build/ci changes should be very similar to vue-next itself

PR #112, opened 2020-02-21
build(dependabot): add config file

vinayakkulkarni:
- [x] Ensures @posva is the reviewer whenever @dependabot opens a new PR for deps upgrade


posva:
Thank you but I prefer not bloating the project with extra config folders that can be avoided (since dependabot options can be configured on its interface)

PR #122, opened 2020-02-26
refactor: make it more readable and self-explanatory

iendeavor:


posva:
can you do a check of the size with something like https://github.com/ai/size-limit ?
I think the previous version minifies and compresses better
I don't really find the new version more readable either

iendeavor:
Both are same size (8.31KB).
In the new version, encode strings are single source of truth and developers have no need to worry about what the encode strings are.

iendeavor:
Thanks for reply.
I forgot to read the contributing guide at first, If you think this commit is no necessary to change, I'll close the commit and implement more valuable commit (wish I could do that) in future.

PR #123, opened 2020-02-29
Refactor error with enum codes

fnlctrl:


src/errors-new.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+interface RouterError {
+  type: ErrorTypes
+  message: string
+  from?: RouteLocation
+  to?: RouteLocation
+}
+
+const ErrorTypeMessages = {
```
posva: I think we need some kind of `logError` method, similar to vue-next. In our case error codes are part of the public api but I think the messages should only be logged in dev mode and just print the error in prod
At this point I wonder if `message` should just be kept empty 🤔

src/errors-new.ts
```diff
@@ -0,0 +1,98 @@
...
+  MatcherLocationNormalized,
+  RouteLocation,
+  RouteLocationNormalized,
+} from './types'
+
+// Using string enums because error codes are exposed to developers
+// and number enums could collide with other error codes in runtime
+export enum ErrorTypes {
+  NO_ROUTE_MATCH = 'NO_ROUTE_MATCH',
+  INVALID_ROUTE_MATCH = 'INVALID_ROUTE_MATCH',
```
posva: This error was removed from the codebase

src/errors-new.ts
```diff
@@ -0,0 +1,98 @@
...
+  MatcherLocation,
+  MatcherLocationNormalized,
+  RouteLocation,
+  RouteLocationNormalized,
+} from './types'
+
+// Using string enums because error codes are exposed to developers
+// and number enums could collide with other error codes in runtime
+export enum ErrorTypes {
+  NO_ROUTE_MATCH = 'NO_ROUTE_MATCH',
```
posva: Let's put `MATCHER_` in front
```suggestion
  MATCHER_NOT_FOUND = 'MATCHER_NOT_FOUND',
```

src/errors-new.ts
```diff
@@ -0,0 +1,98 @@
...
+    from: RouteLocationNormalized,
+    to: RouteLocationNormalized
+  ) {
+    return `Navigation cancelled from "${from.fullPath}" to "${to.fullPath}" with a new \`push\` or \`replace\``
+  },
+}
+
+export function createRouterError<Type extends ErrorTypes>(
+  type: Type,
+  ...messageArgs: Parameters<typeof ErrorTypeMessages[Type]>
```
posva: I think we could have a more generic creation error similar to https://www.typescriptlang.org/play/index.html#code/JYOwLgpgTgZghgYwgAgEoHsCukoFEpTpTIQAekIAJgM7L6HEDeAUMm8mAJ4AOEAXMgDkcQcgA+QgEaDmAX2bNQOeEjoEiAQRLkIVWhmzR6RZC3Yce-ISNbs4A6mCigA5nIVLoKlMagAhbQoaNCwcX1NbNi5eAUFpSORJBydXd2YYTBAEMGB0EGQEKAg4SF8AHlxA3WCDMPUoAD4ACmirXABtQVbBAF0AGmRuOCg4AFtqAQB5UeAwCoGuy1EJAGsITnQYNQYGgEoBSrN2ABsIMBJkAF5kEAgAd22iJt3kOFpcBXMYEybT87XOMhQINhmNqC8IO0Aa9aADNnQelcQSNxlD1jDkHCtq14UMUdQep92EUwJgoPkIGk-hdroViqV6hV6hpmsJBANGK9YgALdaiWS7ZhAA

We could remove the stack entry added by the function call too, similarly to https://github.com/vuejs/vue-router/pull/3047/files#diff-b9b97e885420591ac8b317a611558999R6-R8. Although it's better to check if `stack` exists first. But given that the `stack` property is still not considered stable and the benefit is only removing one stack entry, we don't really need to do it

src/errors-new.ts
```diff
@@ -0,0 +1,98 @@
...
+    type === ErrorTypes.NAVIGATION_CANCELLED ||
+    type === ErrorTypes.NAVIGATION_GUARD_REDIRECT
+  ) {
+    error.from = messageArgs[0]
+    error.to = messageArgs[1]
+  }
+  return error
+}
+
+const propertiesToLog: (keyof RouteLocationNormalized)[] = [
```
posva: I think we can remove the type here and let TS infer it
```suggestion
const propertiesToLog = [
```

src/errors-new.ts
fnlctrl: I added a __DEV__ check to remove it in prod build. `logError` seems unnecessary as we're not catching all errors then print/rethrow them like in core, maybe?

src/errors-new.ts
fnlctrl: I tried to remove the necessity to specify generic when calling, and ended up with this: https://github.com/vuejs/vue-router-next/blob/ef17a9f75fb1fddb9a0bf946e426079e7c25592b/src/errors-new.ts#L36-L40

src/utils/guardToPromiseFn.ts
```diff
@@ -19,9 +19,15 @@ export function guardToPromiseFn(
...
         // TODO: handle callback
-        if (valid === false) reject(new NavigationAborted(to, from))
+        if (valid === false)
+          reject(createRouterError(ErrorTypes.NAVIGATION_ABORTED, { from, to }))
         else if (isRouteLocation(valid)) {
-          reject(new NavigationGuardRedirect(to, valid))
+          reject(
+            createRouterError(ErrorTypes.NAVIGATION_GUARD_REDIRECT, {
+              from: to,
+              to: valid as RouteLocationNormalized, // todo: remove as
```
fnlctrl: here `valid` is just `RouteLocation`, how should we make it normalized?

src/router.ts
```diff
@@ -345,7 +351,13 @@ export function createRouter({
   ) {
     // a more recent navigation took place
     if (pendingLocation !== toLocation) {
-      return triggerError(new NavigationCancelled(toLocation, from), isPush)
+      return triggerError(
+        createRouterError(ErrorTypes.NAVIGATION_CANCELLED, {
+          from: toLocation,
```
posva: isn't it the other way around? `from, to: toLocation`?

src/router.ts
```diff
@@ -397,19 +409,25 @@ export function createRouter({
...
-      if (NavigationGuardRedirect.is(error)) {
+      if (error.type === ErrorTypes.NAVIGATION_GUARD_REDIRECT) {
         // TODO: refactor the duplication of new NavigationCancelled by
         // checking instanceof NavigationError (it's another TODO)
         // a more recent navigation took place
         if (pendingLocation !== toLocation) {
-          return triggerError(new NavigationCancelled(toLocation, from), false)
+          return triggerError(
+            createRouterError(ErrorTypes.NAVIGATION_CANCELLED, {
+              from: toLocation,
```
posva: same here?

src/utils/guardToPromiseFn.ts
posva: We don't need to make it normalised for the redirect. In that case the `to` property can either be `RouteLocation` or `RouteLocationNormalized`. I forgot about it in my previous review 😅

src/errors-new.ts
```diff
@@ -0,0 +1,88 @@
...
+  to: RouteLocationNormalized
+}
+
+type InferErrorType<Type extends ErrorTypes> = Type extends MatcherError['type']
+  ? MatcherError
+  : Type extends NavigationError['type']
+  ? NavigationError
+  : never
+
+const ErrorTypeMessages = __DEV__
```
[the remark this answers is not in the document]
fnlctrl: I tried using the object and only check inside createRouterError, but somehow the compressor didn't strip it out...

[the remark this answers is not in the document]
posva: that's weird, it seems to be like it should like in vue-next. I will take a look

src/router.ts
fnlctrl: oops😂

src/errors-new.ts
```diff
@@ -0,0 +1,92 @@
...
+  currentLocation?: MatcherLocationNormalized
+}
+
+interface NavigationError extends RouterError {
+  type:
+    | ErrorTypes.NAVIGATION_GUARD_REDIRECT
+    | ErrorTypes.NAVIGATION_ABORTED
+    | ErrorTypes.NAVIGATION_CANCELLED
+  from: RouteLocationNormalized
+  to: RouteLocation | RouteLocationNormalized
```
posva: Only the NavigationRedirect has this limitation. We should have an extra type that extends Navigation Error for that one

PR #136, opened 2020-03-15
parseQuery override for end user & test

BboyAkers:
Let me know what you think!


__tests__/router.spec.ts
```diff
@@ -70,6 +69,13 @@ describe('Router', () => {
     )
   })
 
+  it('can allows the end use to override parseQuery', async () => {
+    const customQueryParser = jest.fn()
+    const { router } = await newRouter({ parseQuery: customQueryParser })
+    router.resolve('/foo?bar=baz')
+    expect(customQueryParser).toHaveBeenCalled()
```
posva: Let's call it `parseQuery` so it can just be passed around like this and also let's check it's called with the right parameters
```suggestion
  it('can override parseQuery', async () => {
    const parseQuery = jest.fn()
    const { router } = await newRouter({ parseQuery })
    router.resolve('/foo?bar=baz')
    expect(parseQuery).toHaveBeenCalledWith('?bar=baz')
```

src/router.ts
```diff
@@ -54,6 +54,7 @@ export interface RouterOptions {
   history: RouterHistory
   routes: RouteRecord[]
   scrollBehavior?: ScrollBehavior
+  parseQuery?: typeof originalParseQuery
```
posva: This looks good, let's do the same thing for `stringifyQuery`

src/router.ts
```diff
@@ -65,6 +66,7 @@ export interface Router {
   addRoute(route: RouteRecord): () => void
   removeRoute(name: string): void
   getRoutes(): RouteRecordNormalized[]
+
```
posva: this line can be removed

BboyAkers:
Thanks! I'll start working on it first thing tomorrow!

__tests__/router.spec.ts
```diff
@@ -70,6 +69,13 @@ describe('Router', () => {
     )
   })
 
+  it('can allows the end user to override parseQuery', async () => {
```
posva: let's add another test for `stringifyQuery`

posva:
You have a `lint:fix` command that formats the code: `npm run lint:fix`

BboyAkers:
Updated it all!

BboyAkers:
Done! Thoughts?

PR #178, opened 2020-04-16
fix: update to vue beta 1

killix:
Update to Vue beta 1


posva:
Thank you for this, I had to remove a few other things as well and generate the `yarn.lock` so. I did this locally

killix:
I am not using yarn. Sorry for the forget...

PR #204, opened 2020-04-29
refactor: use named export and add CreateHistoryFunction type

meteorlxy:
Two changes:

### 1. Use named export

Remove `export default` in `history/xxx` dir

### 2. Add `CreateHistoryFunction` type

Just like [CreateAppFunction](https://github.com/vuejs/vue-next/blob/ded92f93b423cda28a40746c1f5fa9bcba56e80d/packages/runtime-core/src/apiCreateApp.ts#L103-L106) in vue-next

### DIFF
--- a/__tests__/history/hash.spec.ts
+++ b/__tests__/history/hash.spec.ts
@@ -1,6 +1,6 @@
 import { JSDOM } from 'jsdom'
-import createWebHashHistory from '../../src/history/hash'
-import createWebHistory from '../../src/history/html5'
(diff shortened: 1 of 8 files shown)


meteorlxy:
I'm developing a propotype of vuepress-next.

As vuepress has both client and server entry, they should use different history creator. So I created a `CreateHistoryFunction` type to declare to history creator type. But I think it's better to be declared in `vue-router` itself.

What's your opinion?

posva:
I added your suggestions at 729721b

PR #206, opened 2020-04-30
feat: allow params number

pikax:
Allow have `param` with number as a value

### DIFF
--- a/__tests__/router.spec.ts
+++ b/__tests__/router.spec.ts
@@ -253,6 +253,16 @@ describe('Router', () => {
     expect(router.currentRoute.value.hash).toBe('#two')
   })
 
+  it('allow param value to be number', async () => {
+    const { router } = await newRouter()
+    router.addRoute({ path: '/r1/:r', name: 'r1', component: components.Bar })
+    expect(router.resolve({ name: 'r1', params: { r: 1 } })).toMatchObject({
+      name: 'r1',
+      path: '/r1/1',
+      params: { r: '1' },
+    })
+  })
+
   it('fails if required params are missing', async () => {
     const { router } = await newRouter()
     expect(() => router.resolve({ name: 'Param', params: {} })).toThrowError(
--- a/__tests__/router.spec.ts
+++ b/__tests__/router.spec.ts
@@ -253,7 +253,7 @@ describe('Router', () => {
     expect(router.currentRoute.value.hash).toBe('#two')
   })
 
-  it('allow param value to be number', async () => {
+  it('cast number params to string', async () => {
     const { router } = await newRouter()
     router.addRoute({ path: '/r1/:r', name: 'r1', component: components.Bar })
     expect(router.resolve({ name: 'r1', params: { r: 1 } })).toMatchObject({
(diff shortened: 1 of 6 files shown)


__tests__/router.spec.ts
```diff
@@ -253,6 +253,16 @@ describe('Router', () => {
     expect(router.currentRoute.value.hash).toBe('#two')
   })
 
+  it('allow param value to be number', async () => {
```
posva, line 256:
```suggestion
  it('cast number params to string', async () => {
```

__tests__/router.spec.ts
```diff
@@ -253,6 +253,16 @@ describe('Router', () => {
...
+    const { router } = await newRouter()
+    router.addRoute({ path: '/r1/:r', name: 'r1', component: components.Bar })
+    expect(router.resolve({ name: 'r1', params: { r: 1 } })).toMatchObject({
+      name: 'r1',
+      path: '/r1/1',
+      params: { r: '1' },
```
posva, line 262: let's not add a new route and use the one named Param
```suggestion
    expect(router.resolve({ name: 'Param', params: { p: 1 } })).toMatchObject({
      name: 'Param',
      path: '/p/1',
      params: { p: '1' },
```

src/router.ts
```diff
@@ -264,7 +264,10 @@ export function createRouter(options: RouterOptions): Router {
       // for same reason TS thinks rawLocation.params can be undefined
       matcher.resolve(
         'params' in rawLocation
-          ? { ...rawLocation, params: encodeParams(rawLocation.params) }
+          ? {
+              ...rawLocation,
+              params: encodeParams(applyToParamsRaw(rawLocation.params)),
```
posva, line 269: A few thoughts for refactoring:

- `encodeParams` could accept numbers
- we could create a copy of `rawLocation` before L256, and add an else at L261 to change the `params` property to an encoded version without numbers, save the decoded version as a local variable `params` to use it later

src/utils/index.ts
```diff
@@ -20,3 +36,9 @@ export function applyToParams(
 
   return newParams
 }
+
+export function applyToParamsRaw(
+  params: RouteParamsRaw | undefined
+): RouteParams | undefined {
+  return applyToParams(x => x.toString(), params)
+}
```
posva, line 44: I think we could write this one as

```suggestion
export let normalizeParams = applyParams.bind(null, paramValue => '' + paramValue)
```

src/router.ts
```diff
@@ -275,7 +278,7 @@ export function createRouter(options: RouterOptions): Router {
     // TODO: normalize params if we accept numbers as raw values
     matchedRoute.params =
       'params' in rawLocation
-        ? rawLocation.params!
+        ? applyToParamsRaw(rawLocation.params)!
```
posva, line 281: we can remove the todo above

src/utils/index.ts
pikax: noticed `let` instead of `const`, is that on purpose? If it is, why? :)

src/router.ts
pikax: `encodeParams` is accepting `string | number` but the `bind` doesn't pick up correct overload, I will just create a const with the `normalized` params and use that across

posva:
I ended up doing something simpler by not allowing number params for the matcher at ef0920a86574bca10836214015c2317ed11a29b7
I still kept you as the author of the commit, I hope that's fine

PR #227, opened 2020-05-08
test(e2e): fix transition e2e specs by increasing css timing

JessicaSachs:
When debugging via the HTML, the transition spec had the right html right before it was failing... so this is simply a matter of timing. By increasing the time the transition takes, we have a better chance of not failing.

I think that because each assertion takes at least 10-15ms. When the transition was set to 500ms, we were cutting it "too close". Sometimes the transition would finish before the assertions did.

Here's the command I used before/after the failures to console.log the body. I couldn't figure out the whole reporter ecosystem so I'm not sure how to bake this into Nightwatch's "afterEach" hooks to make CI debugging better.

```js
// ... assert, waitForElementPresent, etc
.source(result => console.log(result.value))
```

### DIFF
--- a/e2e/specs/transitions.js
+++ b/e2e/specs/transitions.js
@@ -6,7 +6,7 @@ module.exports = {
   '@tags': ['no-headless'],
 
   transitions: function (browser) {
-    const TIMEOUT = 2000
+    const TIMEOUT = 3000
 
     browser
       .url('http://localhost:8080/transitions/')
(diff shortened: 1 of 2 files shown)


JessicaSachs:
The max I saw for an assertion was .18ms on my fast MacBook Pro. There’s
2-3 assertions that need to complete within the transition time frame but
CI might be slower.

We can lower it to 0.75 and see what happens

posva:
Thank you for this Jess!

PR #229, opened 2020-05-09
feat(types): export NavigationGuardNext

cexbrayat:
You currently have to type the `next` arg manaully in a navigation guard

### DIFF
--- a/test-dts/createRouter.test-d.ts
+++ b/test-dts/createRouter.test-d.ts
@@ -1,4 +1,10 @@
-import { createRouter, createWebHistory } from './index'
+import {
+  createRouter,
+  createWebHistory,
+  NavigationGuard,
+  NavigationGuardCallback,
+  RouteLocationNormalized,
+} from './index'
 import { createApp, defineComponent } from 'vue'
 
 const component = defineComponent({})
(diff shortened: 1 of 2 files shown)


test-dts/createRouter.test-d.ts
```diff
@@ -14,5 +20,12 @@ const router = createRouter({
   scrollBehavior(to, from, savedPosition) {},
 })
 
+export const loggedInGuard: NavigationGuard = (
```
posva, line 23: isn't this type redundant? I think the type arguments make more sense with a plain function:


```js
function otherGuard(to: RouteLocationNormalized, from: RouteLocationNormalized, next: NavigationGuardCallback) {}
```

I also think we should rename it to `NavigationGuardNext`, don't do it in the PR to keep it simple but what do you think?

cexbrayat: I think I agree that `NavigationGuard` is not super necessary. `NavigationGuardNext` is probably better indeed 👍 And `RouteLocationNormalized` could be simplified to `RouteLocation`?

posva: I want to but there are a few more types that represent different kinds of route locations. I think I need to do some refactoring when it comes to naming and would definitely welcome some feedback

test-dts/createRouter.test-d.ts
```diff
@@ -14,5 +20,12 @@ const router = createRouter({
...
+  to: RouteLocationNormalized,
+  from: RouteLocationNormalized,
+  next: NavigationGuardCallback
+) => next('/')
+router.beforeEach(loggedInGuard)
```
posva, line 28:
```suggestion
export const loggedInGuard: NavigationGuard = (to, from, next) => next('/')
function beforeGuardFn(
  to: RouteLocationNormalized,
  from: RouteLocationNormalized,
  next: NavigationGuardCallback
) {}

router.beforeEach(loggedInGuard)
router.beforeEach(beforeGuardFn)
```

PR #258, opened 2020-05-19
Document useRoute and useRouter

jeffreybiles:
A documentation partial solution to https://github.com/vuejs/vue-router-next/issues/257 - I'll look at creating a warning solution as well when I get time.

I'm not 100% sure whether the last sentence is technically correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,24 @@ The current codebase has most of the existing features on Vue Router v3.x and is
 
 Since the library is still unstable **and because we want feedback** on bugs and missing features, **it will probably go through a few breaking changes**.
 
+## Accessing Route and Routing in the setup function
+
+The new `useRoute` and `useRouter` functions are used to access the route and router in the setup function.
+
+```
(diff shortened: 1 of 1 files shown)


posva:
Thank you but I rather directly add the warnings I mentioned at https://github.com/vuejs/vue-router-next/issues/257#issuecomment-630706991

jeffreybiles:
What about this PR without the final line?

I think `useRoute` and `useRouter` should be documented somewhere - I had to source dive to find them today, and only found them after being tipped by another GitHub issue.

posva:
It's true it makes sense to add a note about them. I will add a note, it should compare it to `$route` and `$router` instead.

jeffreybiles:
Sounds good.  Thank you!

posva:
I didn't make sense to just add a note about those 2 functions so I added a note about the examples where people can find usage. The proper documentation will exist in a different place

PR #279, opened 2020-05-29
test: add test for route not matched

pikax:
Warns when the route is not found

### DIFF
--- a/__tests__/warnings.spec.ts
+++ b/__tests__/warnings.spec.ts
@@ -174,4 +174,17 @@ describe('warnings', () => {
     await expect(router.push({ path: '/foo' })).resolves.toBe(undefined)
     expect('"/foo" is a Promise instead of a function').toHaveBeenWarned()
   })
+
+  it('warns if no route matched', async () => {
+    const router = createRouter({
+      history: createMemoryHistory(),
+      routes: [{ path: '/', name: 'a', component }],
+    })
+
+    await expect(router.push('/foo')).resolves.toBe(undefined)
+    expect(`No route matched for "/foo"`).toHaveBeenWarned()
+
+    await expect(router.push({ path: '/foo2' })).resolves.toBe(undefined)
+    expect(`No route matched for "/foo2"`).toHaveBeenWarned()
+  })
 })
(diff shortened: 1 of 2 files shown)


__tests__/warnings.spec.ts
```diff
@@ -27,6 +27,7 @@ describe('warnings', () => {
     })
     router.resolve({ path: '/', params: { p: 'p' } })
     expect('Path "/" was passed with params').toHaveBeenWarned()
+    expect(`No route matched`).toHaveBeenWarned()
```
posva, line 30: let's avoid the other warning instead

```suggestion
    router.resolve({ path: '/p', params: { p: 'p' } })
    expect('Path "/p" was passed with params').toHaveBeenWarned()
```

__tests__/warnings.spec.ts
```diff
@@ -135,6 +136,7 @@ describe('warnings', () => {
 
     await expect(router.push('//not-valid')).resolves.toBe(undefined)
     expect('cannot start with multiple slashes').toHaveBeenWarned()
+    expect(`No route matched`).toHaveBeenWarned()
```
posva, line 139: same as above, let's add a route `//not-valid`, although it doesn't make much sense

__tests__/warnings.spec.ts
```diff
@@ -145,6 +147,7 @@ describe('warnings', () => {
 
     await expect(router.push({ path: '//not-valid' })).resolves.toBe(undefined)
     expect('cannot start with multiple slashes').toHaveBeenWarned()
+    expect(`No route matched`).toHaveBeenWarned()
```
posva, line 150: same as above

src/router.ts
```diff
@@ -290,6 +294,13 @@ export function createRouter(options: RouterOptions): Router {
     }
 
     let matchedRoute = matcher.resolve(matcherLocation, currentLocation)
+    if (matchedRoute.matched.length === 0) {
```
posva, line 297:
```suggestion
    if (__DEV__ && !matchedRoute.matched.length) {
```

__tests__/warnings.spec.ts
pikax: not adding the route `//not-valid`, because the matcher won't be able to match it, I think if we add it will confuse maintainers in the future.

PR #343, opened 2020-07-01
feat(guards): allow guards to return a value instead of calling next

posva:
Implements https://github.com/vuejs/rfcs/pull/187


ycmjason:
haha nice! using `.length` just like I suggested isn't it? Lovely one! Cheers!

ycmjason:
so good!!!! can't wait to use it!

PR #344, opened 2020-07-02
refactor: refactor router-view + fix keep-alive route guard this context

yyx990803:
- Refactored `<router-view>` implementation
- Upgrading to `vue@3.0.0-beta.18`
  - fixes navigation guard `this` when route component is inside `<keep-alive>`
  - passing component vnode via `<router-view>` slot gets rid of the need to pass `props`, so the user no longer needs to bind it. This simplifies `<router-view>` slot usage with `<transition>/<keep-alive>` to:

    ```html
    <router-view v-slot="{ Component }">
      <transition name="fade" mode="out-in">
        <keep-alive>
          <component :is="Component"></component>
        </keep-alive>
      </transition>
    </router-view>
    ```

    [RFC](https://github.com/vuejs/rfcs/pull/160) needs to be updated accordingly.

  - also added a warning when detecting old usage (`<router-view>` directly inside `<transition>/<keep-alive>`)

### DIFF
--- a/src/RouterView.ts
+++ b/src/RouterView.ts
@@ -4,18 +4,20 @@ import {
   provide,
   defineComponent,
   PropType,
-  computed,
   ref,
   ComponentPublicInstance,
   VNodeProps,
+  getCurrentInstance,
+  computed,
 } from 'vue'
-import { RouteLocationNormalizedLoaded, RouteLocationNormalized } from './types'
+import { RouteLocationNormalized, RouteLocationNormalizedLoaded } from './types'
 import {
   matchedRouteKey,
   viewDepthKey,
   routeLocationKey,
 } from './injectionSymbols'
 import { assign } from './utils'
+import { warn } from './warning'
 
 export interface RouterViewProps {
   name?: string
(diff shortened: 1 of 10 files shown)


src/RouterView.ts
```diff
@@ -34,80 +36,66 @@ export const RouterViewImpl = defineComponent({
...
-
-      return children
-        ? children
-        : Component
-        ? h(Component, componentProps)
-        : null
+      return (
+        // pass the vnode to the slot as a prop.
+        // h and <component :is="..."> both accept vnodes
+        (slots.default && slots.default({ Component: component })) || component
```
yyx990803, line 97: Note here we are passing `component`, which is a vnode, to the slot as `Component`. Changes in `vue@3.0.0-beta.18` allows `<component :is="">` and `h` to accept an existing vnode which makes it possible.

posva, line 97: I think here we could just use a ternary:

```suggestion
        slots.default ? slots.default({ Component: component }) : component
```

posva: Aah, I didn't see the parentheses correctly. But I think it would make more sense to use the slot if there is one without falling back to `component`. This way the user could hide it if they wanted to

yyx990803: Yeah I just followed the original logic. We should probably use a ternary indeed.

PR #360, opened 2020-07-06
fix(router): remove warning whe no match found

### LINKED ISSUE #359: Provide a way to disable warnings for "No match found" when testing
When adding tests to a component using the router, one might need to provide a mock router. I currently use:

```typescript
const mockRouter = createRouter({ history: createWebHistory(), routes: [] });
describe('Navbar.vue', () => {
  test('should ...', () => {
    const navbar = mount(Navbar, {
      global: {
        plugins: [mockRouter],
        stubs: {
          RouterLink: RouterLinkStub
        }
      }
    });
    // ...
  });
});
```

It works perfectly, but the router emits a warning:
```
console.warn node_modules/vue-router/dist/vue-router.cjs.js:225
    [Vue Router warn]: No match found for location with path "/"
```

As most tests are using the same code, the warning is emitted a lot and pollutes the test output.

Until VTU or the router itself provides built-in support for testing (if ever, see https://github.com/vuejs/vue-test-utils-next/issues/152 for curious readers), do you think we could guard this warning with a setting to hide it? I can open a PR if you provide some guidance (and think this is useful of course).

cexbrayat:
Fixes #359 by removing the warning

Let me know if that's good enough, or if you prefer another to disable the warning only for tests.

### DIFF
--- a/__tests__/router.spec.ts
+++ b/__tests__/router.spec.ts
@@ -232,7 +232,6 @@ describe('Router', () => {
     await router.push('/me-neither')
     expect(router.currentRoute.value).toMatchObject({ matched: [] })
     expect(spy).toHaveBeenCalledTimes(1)
-    expect('No match found').toHaveBeenWarnedTimes(2)
   })
 
   it('casts number params to string', async () => {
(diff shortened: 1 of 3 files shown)


posva:
Thanks for this but see https://github.com/vuejs/vue-router-next/issues/359#issuecomment-683654540

PR #361, opened 2020-07-06
repro(types): components with required props

cexbrayat:
This is a repro to illustrate a compilation error with components using required props.

With a route like `{ path: '/', component, props: true }`, if the component has a required prop, then the TS compilation fails with:
```typescript
TS2322: Type 'ComponentPublicInstanceConstructor<ComponentPublicInstance<{ id: string; } & {}, {}, { users: readonly { name: string; }[]; }, {}, {}, Record<string, any>, VNodeProps, ComponentOptionsBase<{ id: string; } & {}, unknown, { users: readonly { name: string; }[]; }, {}, ... 4 more ..., string>>> & ComponentOptionsBase<.....' is not assignable to type 'RawRouteComponent'.
  Type 'ComponentPublicInstanceConstructor<ComponentPublicInstance<{ id: string; } & {}, {}, { users: readonly { name: string; }[]; }, {}, {}, Record<string, any>, VNodeProps, ComponentOptionsBase<{ id: string; } & {}, unknown, { users: readonly { name: string; }[]; }, {}, ... 4 more ..., string>>> & ComponentOptionsBase<.....' is not assignable to type 'FunctionalComponent<any, Record<string, any>>'.
    Type 'ComponentPublicInstanceConstructor<ComponentPublicInstance<{ id: string; } & {}, {}, { users: readonly { name: string; }[]; }, {}, {}, Record<string, any>, VNodeProps, ComponentOptionsBase<{ id: string; } & {}, unknown, { users: readonly { name: string; }[]; }, {}, ... 4 more ..., string>>> & ComponentOptionsBase<.....' provides no match for the signature '(props: any, ctx: SetupContext<Record<string, any>>): any'.
```

Update: I hoped to reproduce with a tsd test, but it does not fail, so there might be something fishy with the tsd setup?

### DIFF
--- a/e2e/modal/index.ts
+++ b/e2e/modal/index.ts
@@ -1,6 +1,14 @@
 import { createRouter, createWebHistory, useRoute } from '../../src'
 import { RouteComponent, RouteLocationNormalizedLoaded } from '../../src/types'
-import { createApp, readonly, ref, watchEffect, computed, toRefs } from 'vue'
+import {
+  createApp,
+  readonly,
+  ref,
+  watchEffect,
+  computed,
+  toRefs,
+  defineComponent,
+} from 'vue'
 
 const users = readonly([
   { name: 'John' },
@@ -94,7 +102,7 @@ const About: RouteComponent = {
   },
 }
 
-const UserDetails: RouteComponent = {
+const UserDetails = defineComponent({
   template: `<div>
     <h1>User #{{ id }}</h1>
     <p>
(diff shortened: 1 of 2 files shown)


cexbrayat:
Hmm, this is not failing the tsd tests, weird... 
It does fail in my project, and you can reproduce in the e2e test of the modal (but the test still succeeds even if it does not compile). I pushed another commit

PR #373, opened 2020-07-16
test(__tests__/RouterLink.spec): add not match repeated params test

uuoocckk-983:


__tests__/RouterLink.spec.ts
```diff
@@ -515,6 +530,15 @@ describe('RouterLink', () => {
     expect(wrapper.find('a')!.className).toBe('')
   })
 
+  it('is not active with not match repeated params', async () => {
```
posva, line 533:
```suggestion
  it('is not active with different repeated params', async () => {
```

Issue #404, opened 2020-08-04
Stack overflow when named location redirect to the other by path-string


LittleIQ:
### Version
4.0.0-beta.5

### Reproduction link
[https://jsfiddle.net/yvdk49z8/1/](https://jsfiddle.net/yvdk49z8/1/)

### Steps to reproduce
use a named location redirect to the other by path-string

### What is expected?
Same behaviour between named location redirect vs unamed location redirect

### What is actually happening?
When named location redirect to the other one by path-string, only path-string is changed in newTargetLocation, but resolve path by old name first.

posva:
As a workaround until a new version is released, use a named route `{ name: 'child' }` in the `redirect` option.

PR #405, opened 2020-08-04
fix: use assign to update newTargetLocation (#404)

LittleIQ:


posva:
Thank you, I submitted a fix already to properly keep only what is needed!

PR #408, opened 2020-08-06
Improve Vetur component data

octref:


vetur/tags.json
```diff
@@ -1,10 +1,10 @@
 {
   "router-view": {
     "attributes": ["name", "route"],
-    "description": "Component that renders the matched component for the current location. Components rendered by `<router-view>` can also contain their own `<router-view>` to render nested routes."
+    "description": "Component that renders the matched component for the current location. Components rendered by `<router-view>` can also contain their own `<router-view>` to render nested routes.\n\n[API Reference](https://router.vuejs.org/api/#router-link)"
   },
   "router-link": {
     "attributes": ["to", "activeClass","exactActiveClass", "custom"],
-    "description": "Component that renders an `<a>` with the correct `href` attribute and click listeners to trigger a local navigation when clicked. Can also customize its rendering by providing the `custom` prop and using its `v-slot` API."
+    "description": "Component that renders an `<a>` with the correct `href` attribute and click listeners to trigger a local navigation when clicked. Can also customize its rendering by providing the `custom` prop and using its `v-slot` API.\n\n[API Reference](https://router.vuejs.org/api/##router-view)"
```
posva, line 8: there is a double `#`

```suggestion
    "description": "Component that renders an `<a>` with the correct `href` attribute and click listeners to trigger a local navigation when clicked. Can also customize its rendering by providing the `custom` prop and using its `v-slot` API.\n\n[API Reference](https://router.vuejs.org/api/#router-view)"
```

vetur/tags.json
```diff
@@ -1,10 +1,10 @@
 {
   "router-view": {
     "attributes": ["name", "route"],
-    "description": "Component that renders the matched component for the current location. Components rendered by `<router-view>` can also contain their own `<router-view>` to render nested routes."
+    "description": "Component that renders the matched component for the current location. Components rendered by `<router-view>` can also contain their own `<router-view>` to render nested routes.\n\n[API Reference](https://router.vuejs.org/api/#router-link)"
   },
   "router-link": {
     "attributes": ["to", "activeClass","exactActiveClass", "custom"],
```
octref, line 7: Why are you not including all attributes for `router-link`?

posva: I forgot `ariaCurrentValue` because I added it later I think

posva: all the other props have been removed

octref: 🤦‍♂️ I was comparing it against old Vue router docs. Do you have a new one available? Can't find it. Also if they would be in different sites I should link to the one for vue-router-next.
(BTW I wasn't aware, are all attributes renamed to camelCase?)

posva: There isn't an official link yet, so I think we can add that later.
props can be passed both ways but by what you are saying I understand I should provide the kebab-case version for autocompletion 😄

octref: @posva I can add functionality to handle camelCase to kebab-case as well. No worries, just need to do a bit of changes in Vetur side.

posva: I think autocompleting on kebab-case is the way to go because it works on different contexts whereas camelCase doesn't. We should just use kebab-case in these files

vetur/attributes.json
```diff
@@ -13,6 +13,10 @@
     "type": "boolean",
     "description": "Setting replace prop will call `router.replace()` instead of `router.push()` when clicked, so the navigation will replace the current history entry.\n\n[API Reference](https://router.vuejs.org/api/#replace)"
   },
+  "custom": {
+    "type": "boolean",
+    "description": "Whether `RouterLink` should not wrap its content in an `a` tag."
```
posva, line 18: I think using the kebab case annotation plus wrapping with `<>` will make it more familiar:
```suggestion
    "description": "Whether `<router-link>` should not wrap its content in an `<a>` tag."
```

octref:
> remove the links

posva:
yes, because we don't have the right links so pointing to outdated ones could be confusing

octref:
I didn't notice the contributing guideline. I squashed and gave a new commit message.

octref:
@posva That's up to you — anytime you want. Thanks for reviewing 👍

PR #421, opened 2020-08-14
fix(types): allow components defined via defineComponent

### LINKED ISSUE #420: Types error when component is defined via defineComponent
### Version
4.0.0-beta.6

### Reproduction link
[https://codesandbox.io/s/sleepy-almeida-y5m5l?file=/src/index.ts](https://codesandbox.io/s/sleepy-almeida-y5m5l?file=/src/index.ts)

### Steps to reproduce
```ts
import { createApp, h, defineComponent } from 'vue';
import { createWebHistory, createRouter, RouterView } from 'vue-router';

export const Home = defineComponent({
  props: {
    test: {
      type: String,
      required: true,
    },
  },
  setup() {
    return {
      title: 'homepage'
    }
  },
  render() {
    return h('div', `${this.title}: ${this.test}`)
  },
});
const router = createRouter({
  history: createWebHistory(),
  routes: [{
    path: '/home/:test',
    name: 'home',
    props: true,
    component: Home, // types error in here
  }, {
    path: '/:catchAll(.*)',
    redirect: { name: 'home', params: { test: 'redirect' } },
  }],
});

const app = createApp({
  render() {
    return h(
      RouterView,
    );
  }
});

app.use(router);
app.mount('#app');

```
the error message is 
```
Type 'ComponentPublicInstanceConstructor<ComponentPublicInstance<{ test: string; } & {}, { title: string; }, {}, {}, {}, Record<string, any>, VNodeProps & AllowedComponentProps & ComponentCustomProps, ComponentOptionsBase<...>>> & ComponentOptionsBase<...> & { ...; } & ThisType<...>' is not assignable to type 'RawRouteComponent'.
  Type 'ComponentPublicInstanceConstructor<ComponentPublicInstance<{ test: string; } & {}, { title: string; }, {}, {}, {}, Record<string, any>, VNodeProps & AllowedComponentProps & ComponentCustomProps, ComponentOptionsBase<...>>> & ComponentOptionsBase<...> & { ...; } & ThisType<...>' is not assignable to type 'FunctionalComponent<any, {}>'.
    Type 'ComponentPublicInstanceConstructor<ComponentPublicInstance<{ test: string; } & {}, { title: string; }, {}, {}, {}, Record<string, any>, VNodeProps & AllowedComponentProps & ComponentCustomProps, ComponentOptionsBase<...>>> & ComponentOptionsBase<...> & { ...; } & ThisType<...>' provides no match for the signature '(props: any, ctx: SetupContext<{}>): any'.ts(2322)
```

### What is expected?
the right types.

types error.

posva:
Close #420
Close #361

I'm not sure about this one, specially the one way I extract PublicAPIComponent because If I don't add he cast to `Component`, `h()` yields a type error. @yyx990803 do you think it makes sense to expose `PublicAPIComponent`

Maybe you have a better idea @pikax, @cexbrayat ?


cexbrayat:
@posva I have no better idea 🤔 We also struggled in VTU to get it right (and we might still have some issues). If @yyx990803 agrees, I think it would be great if Vue exposed a directly usable type 👍 The shim and libs would be simplified

src/types/index.ts
```diff
@@ -161,7 +161,12 @@ export interface RouteLocationNormalized extends _RouteLocationBase {
   matched: RouteRecordNormalized[] // non-enumerable
 }
 
-export type RouteComponent = Component
+type PublicAPIComponent = Exclude<
+  Parameters<typeof defineComponent>[0]['components'],
+  undefined
+>[string]
+
+export type RouteComponent = PublicAPIComponent
```
pikax, line 169: I would do:
```ts
export type RouteComponent = Component | ReturnType<typeof defineComponent>
```

which will have all possible cases when using defineComponent

posva: I also went too complicated by looking up the Parameters of `defineComponent` 😆

Senasiko: ReturnType of overload functions just use the last overload for inference, look at [this issue](https://github.com/Microsoft/TypeScript/issues/24275#issuecomment-390701982). So may we need to find another way？

pikax: @Senasiko can you provide an overload where this is failing?

Senasiko:
```
export const Home = defineComponent({
  setup() {
    return {}
  },
  props: {
    test: {
      type: String,
      required: true,
    }
  },
  render() {},
});

const router = createRouter({
  history: createWebHistory(),
  routes: [{
    path: '/home/:test',
    name: 'home',
    props: true,
    component: Home,
  }],
});
```
If `defineComponent`'s param has `props` option, I will get types error in `component: Home,`, even if I write `declare type RouteComponent = Component | ReturnType<typeof defineComponent>;` in `vue-router.d.ts`

pikax: Having the generic constructor should be enough:
```ts
declare type RouteComponent =
  | Component
  | { new (): ComponentPublicInstance }
```

PR #446, opened 2020-08-26
fix(router-view): reuse saved instances in different records

### LINKED ISSUE #444: beforeRouteUpdate not work when different routes route to same Component
### Version
4.0.0-beta.7

### Reproduction link
[https://codesandbox.io/s/vue-router-next-beforerouteupdate-qpclb](https://codesandbox.io/s/vue-router-next-beforerouteupdate-qpclb)

### Steps to reproduce
* Have different routes points to the same Component
* Register `beforeRouteLeave`, `beforeRouteEnter` and `beforeRouteUpdate`
* Switch between `router-link`, you would not receive all navigation event

### What is expected?
1. Click on `1`, shows "leave /app/1" and "enter /app/1"
2. Click on `2`, shows "update /app/2"
3. Click on `Home`, shows "leave /" and "enter /"

1. Click on `1`, shows "leave /app/1" and "enter /app/1"
2. Click on `2`, nothing shows
3. Click on `Home`, shows "enter /"

posva:
Close #444

Still worth adding edge cases

- Keep alive
- Dynamic names

### DIFF
--- a/__tests__/guards/onBeforeRouteLeave.spec.ts
+++ b/__tests__/guards/onBeforeRouteLeave.spec.ts
@@ -13,58 +13,10 @@ const component = {
 }
 
 describe('onBeforeRouteLeave', () => {
-  it('invokes with the component context', async () => {
-    expect.assertions(2)
-    const spy = jest
-      .fn()
-      .mockImplementationOnce(function (this: any, to, from, next) {
-        expect(typeof this.counter).toBe('number')
-        next()
-      })
-    const WithLeave = defineComponent({
-      template: `text`,
-      // we use data to check if the context is the right one because saving `this` in a variable logs a few warnings
-      data: () => ({ counter: 0 }),
-      setup() {
-        onBeforeRouteLeave(spy)
-      },
-    })
-
-    const router = createRouter({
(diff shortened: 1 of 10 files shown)


posva:
The problem with this fix is that it doesn't fix update and leave guards inside `setup` because `setup` doesn't get re-executed and I don't have a way to tell if and which component instances are reused. In that one needs to use a `key` attribute to force the `setup` to get executed again. We could use the `$route.path`:

```html
<router-view :key="$route.path" />
```

but that would also recreate routes with different params. A most appropriate solution would be using the record and the name of the view (`default` when there is only one view). The problem is this information is not accessible right now outside of RouterView because one needs to know the depth:

```html
<!-- path is here the record's path e.g. /users/:id -->
<router-view name="named" :key="'named_' + $route.matched[depth].path" />
<router-view :key="'default_' + $route.matched[depth].path" />
```

This information could be exposed through the `v-slot` when having the need to reuse the same component for different routes, something like `key` in addition to `Component` and `route`.

posva:
At the end I manage to find a solution that reuses view instances so I went with that instead

PR #449, opened 2020-09-01
test: delete duplicate test case

ruige24601:
Two test case are same:
```
  it('passes params as props with props: true', async () => {
    const { wrapper, route } = await factory(routes.withParams)
    expect(wrapper.html()).toBe(`<div>User: 1</div>`)
    await route.set({
      ...routes.withParams,
      params: { id: '4' },
    })
    expect(wrapper.html()).toBe(`<div>User: 4</div>`)
  })

 it('passes params as props with props: true', async () => {
    const { wrapper, route } = await factory(routes.withParams)

    expect(wrapper.html()).toBe(`<div>User: 1</div>`)

    await route.set({
      ...routes.withParams,
      params: { id: '4' },
    })
    expect(wrapper.html()).toBe(`<div>User: 4</div>`)
  })
```

### DIFF
--- a/__tests__/RouterView.spec.ts
+++ b/__tests__/RouterView.spec.ts
@@ -288,18 +288,6 @@ describe('RouterView', () => {
     expect(wrapper.html()).toBe(`<div>User: 4</div>`)
   })
 
-  it('passes params as props with props: true', async () => {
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

PR #476, opened 2020-09-21
fix(router): allow `options.routes` is not provided

evillt:
Ref
https://github.com/vuejs/vue-router/blob/ccbd917458e73a4eeff9158674893b16de7f79c1/src/index.js#L47


posva:
Thanks but `routes` are **required** in options

PR #490, opened 2020-09-22
Optimize: enum should be const

dolymood:


src/errors.ts
```diff
@@ -34,7 +34,7 @@ export interface MatcherError extends Error {
  * Enumeration with all possible types for navigation failures. Can be passed to
  * {@link isNavigationFailure} to check for specific failures.
  */
-export enum NavigationFailureType {
+export const enum NavigationFailureType {
```
posva, line 37: Since this enum is exposed to the user, I doubt it's a good idea to make it const. It needs to be exposed to JS users as well

posva:
Thanks for this although I'm not sure we want to make them const if they are all meant to be used by the end JS user as well as TS users

dolymood:
Thanks for you CR.

PR #506, opened 2020-09-30
refactor(matcher): add error message for v3 catchall

miyaoka:
Got an error `Route "*" should be "/*".` when I set `path: '*'`.
However it still doesn't work (I didn't notice new catchall rule).

So added new error message for catchall.

### DIFF
--- a/__tests__/matcher/pathParser.spec.ts
+++ b/__tests__/matcher/pathParser.spec.ts
@@ -13,6 +13,16 @@ describe('Path parser', () => {
       expect(tokenizePath('')).toEqual([[]])
     })
 
+    it('v3 catchAll', () => {
+      expect(() => tokenizePath('*')).toThrowError(
+        `Catch all routes (/*) must now be defined using a parameter with a custom regex: /:catchAll(.*)`
+      )
+    })
+
+    it('not start with /', () => {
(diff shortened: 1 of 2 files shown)


posva (changes requested):
Thank you but this should be a DEV only warning at the matcher level when adding a route that is **exactly** `'*'` because it doesn't start with a `/` so it's still fine to warn against it and it could help people migrating. Other paths like `/*` or `/path-*` are technically valid so they shouldn't warn

miyaoka:
`/*` is valid?

I got error `Route "*" should be "/*"`, so I changed `'*'` to `/*`. then got warn `[Vue Router warn]: No match found for location with path`. CatchAll doesn't work.

posva:
`/*` will match the `*` character, it no longer has a special meaning unless put after a param

miyaoka:
Ahhhh, I finally understand. Thank you.
I'll fix it to show warning instead of error.

posva:
I'm going to take over this one because I want to release a new version 🙂 
Thank you!

miyaoka:
OK! Thank you!

posva:
Merged at 14e1eb96485f74669f582a87f522d3b13b567c9c

I couldn't push to the branch, maybe you unticked the option on the right that allows maintainers to edit 🤔 I still kept you as the author though!

PR #510, opened 2020-10-04
docs: add netlify section for history mode rewrites (#510) [skip ci]

CodingCarter:
[Read original post here](https://github.com/vuejs/vuepress/issues/457#issuecomment-463745656)

### DIFF
--- a/docs/guide/essentials/history-mode.md
+++ b/docs/guide/essentials/history-mode.md
@@ -151,14 +151,13 @@ Add this to your `firebase.json`:
 
 ### Netlify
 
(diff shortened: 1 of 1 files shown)


posva (changes requested):
I think it's easier to use a `_redirects` and also add a link to their docs for more advanced usage: https://docs.netlify.com/routing/redirects/rewrites-proxies/#history-pushstate-and-single-page-apps

posva (approved):
Thank you! I will do a couple of tweaks locally

PR #511, opened 2020-10-04
docs: add `$` in front of `route`

CodingCarter:
Using regular `route` does not work within the template.


CodingCarter:
Nvm, you used a slot.

PR #522, opened 2020-10-10
docs: improve scroll-behavior.md (#522) [skip ci]

MostafaSalama:
fixing the syntax in scroll behavior function, the `el: '#main'` should be in a comment and the element **el** should be in the returned object.


posva (approved):
Thanks!

PR #523, opened 2020-10-11
chore: new lines in css [skip ci]

CodingCarter:


docs/.vitepress/theme/components/CarbonAds.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-  <div class="carbon-ads" ref="el"></div>
+  <div class="carbon-ads" ref="el" />
 </template>
 
 <script>
-import { onMounted, ref } from 'vue'
+import { ref, onMounted } from 'vue'
```
posva, line 6: this is not necessary

docs/.vitepress/theme/components/CarbonAds.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-  <div class="carbon-ads" ref="el"></div>
+  <div class="carbon-ads" ref="el" />
```
posva, line 2: same here

docs/.vitepress/theme/components/BuySellAds.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="bsa-cpc-wrapper">
-    <div class="bsa-cpc"></div>
+    <div class="bsa-cpc" />
```
posva, line 3: I rather keep the old version

posva (changes requested):
Thanks!

docs/.vitepress/theme/components/CarbonAds.vue
CodingCarter: Although these may not be necessary, it is strongly recommended for you to self close empty elements, [according to the Vue 2 Style Guide section on the topic](https://vuejs.org/v2/style-guide/#Self-closing-components-strongly-recommended).  The style guide is most likely applicable to Vue 3 as it has the same behavior for self closing components.  I think that the official Vue Router should probably uphold to the official Vue style guide.

docs/.vitepress/theme/components/CarbonAds.vue
CodingCarter: Yep, I figured that it would be better to swap the order, as it makes the imports match the two most common import order conventions: alphabetically and by order of use.  I'll remove it, though.

docs/.vitepress/theme/components/CarbonAds.vue
posva: It's for components not elements. Thanks

CodingCarter: Sorry

PR #562, opened 2020-11-02
docs:Use v-bind for router-link on the API.

Alanscut:


posva:
Thank you, I don't think this is necessary given how early we introduce shorthands in Vue docs: https://v3.vuejs.org/guide/template-syntax.html#v-bind-shorthand and that we recommend them everywhere

PR #605, opened 2020-11-22
docs(guide): update index.md for more comprehensive docs

JulianRomana:
People might think that a `VueRouter` instance is being exported from `vue-router`, adding this line can allow user building app faster and not loosing time on some details


JulianRomana:
I'm opened for discussion if you have any advises or whatever !

posva:
Thank you but this is correct already. The variable comes from the Umd build

PR #642, opened 2020-12-09
docs: remove duplicate `addRoute` description

jessuni:


posva:
thanks but the function has two different signatures!

PR #661, opened 2020-12-16
docs: use chunks link

Alanscut:


posva:
I find the original link to be more appropriate

Alanscut:
Although Dynamic Imports link articles are more comprehensive, the link content of Chunks is easier to understand named chunk.😀

posva:
I don't think so: the chunks one is an advanced section of the guide (under the hood) and most users don't need to know about chunks. Let's keep it this way, in the future, I might change the link to something more comprehensive.

PR #677, opened 2020-12-25
docs: add imports

dsine-de:
Added information about how to import the composition API functions (useRouter & useRoute).


posva (approved):
Thanks!

PR #680, opened 2020-12-27
fix(matcher): clear customRe after consuming buffer

### LINKED ISSUE #679: Custom regex leaks to subsequent parameters without regex
### Version
4.0.1

### Reproduction link
[https://paths.esm.dev/?p=AAMeJbiAwB4U7BTA0AGBCQmoAEUduADagKA.#](https://paths.esm.dev/?p=AAMeJbiAwB4U7BTA0AGBCQmoAEUduADagKA.#)

### Steps to reproduce
Define a path with two parameters, the first using a custom a regex and the second without.

### What is expected?
The regex is applied only for the first parameter, the second parameter uses the default regex.

The custom regex is applied to both parameters.

dneuschaefer-rube:
Clear customRe after consuming buffer to avoid re-use for subsequent parameters.

close #679

### DIFF
--- a/__tests__/matcher/pathParser.spec.ts
+++ b/__tests__/matcher/pathParser.spec.ts
@@ -147,6 +147,29 @@ describe('Path parser', () => {
       ])
     })
 
+    it('param custom re followed by param without regex', () => {
+      expect(tokenizePath('/:one(\\d+)/:two')).toEqual([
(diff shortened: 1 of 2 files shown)


posva (approved):
Perfect! 🙌
Thank you!

PR #682, opened 2020-12-29
feat(warn): warn defineAsyncComponent usage in routes

### LINKED ISSUE #627: Warn against `defineAsyncComponent` usage in `routes`
```
createRouter({
routes: [{ path: '/', component: defineAsyncComponent(() => import('...')) }]
})
```

The code above should warn and hint the user to not use `defineAsyncComponent

leemove:
Close #627

### DIFF
--- a/__tests__/lazyLoading.spec.ts
+++ b/__tests__/lazyLoading.spec.ts
@@ -3,7 +3,7 @@ import { createRouter, createMemoryHistory } from '../src'
 import { RouterOptions } from '../src/router'
 import { RouteComponent } from '../src/types'
 import { ticks } from './utils'
-import { defineAsyncComponent, FunctionalComponent, h } from 'vue'
+import { FunctionalComponent, h } from 'vue'
 import { mockWarn } from 'jest-mock-warn'
 
 function newRouter(options: Partial<RouterOptions> = {}) {
@@ -18,19 +18,6 @@ function createLazyComponent() {
 
   return {
     component: jest.fn(() => promise.then(() => ({} as RouteComponent))),
-    promise,
-    resolve,
-    reject,
-  }
-}
-
-function createLazyAsyncComponent() {
-  const [promise, resolve, reject] = fakePromise()
-
-  return {
-    component: jest.fn(() =>
-      promise.then(() => defineAsyncComponent(() => Promise.resolve({})))
-    ),
     promise,
     resolve,
     reject,
@@ -52,19 +39,6 @@ describe('Lazy Loading', () => {
     resolve()
 
     await p
-    expect(router.currentRoute.value).toMatchObject({
-      path: '/foo',
-      matched: [{}],
-    })
-  })
-
-  it('works with async components', async () => {
-    const { component, resolve } = createLazyAsyncComponent()
-    const { router } = newRouter({
-      routes: [{ path: '/foo', component }],
-    })
-    resolve()
-    await router.push('/foo')
     expect(router.currentRoute.value).toMatchObject({
       path: '/foo',
       matched: [{}],
--- a/__tests__/lazyLoading.spec.ts
+++ b/__tests__/lazyLoading.spec.ts
@@ -3,7 +3,7 @@ import { createRouter, createMemoryHistory } from '../src'
 import { RouterOptions } from '../src/router'
 import { RouteComponent } from '../src/types'
 import { ticks } from './utils'
-import { FunctionalComponent, h } from 'vue'
+import { defineAsyncComponent, FunctionalComponent, h } from 'vue'
 import { mockWarn } from 'jest-mock-warn'
 
 function newRouter(options: Partial<RouterOptions> = {}) {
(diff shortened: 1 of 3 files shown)


__tests__/lazyLoading.spec.ts
```diff
@@ -45,6 +58,19 @@ describe('Lazy Loading', () => {
     })
   })
 
+  it('works with async components', async () => {
```
posva, line 61: the test only needs to check if a warning is raised

src/navigationGuards.ts
```diff
@@ -253,6 +253,13 @@ export function extractComponentsGuards(
           )
           let promise = rawComponent
           rawComponent = () => promise
+        } else if ((rawComponent as any)['__asyncLoader']) {
+          let asyncComponent = rawComponent
+          rawComponent = () => Promise.resolve(asyncComponent)
```
posva, line 258: I think this can be removed

src/navigationGuards.ts
```diff
@@ -253,6 +253,13 @@ export function extractComponentsGuards(
...
+          warn(
+            `Async component in record with path "${record.path}" can not be used in routes. ` +
```
posva, line 260: I would prefer a warning that points what to do:

```js
`Component "${name}" in record with path "${record.path}" is defined ` +
              `using "defineAsyncComponent()". ` +
              `Write "() => import('./MyPage.vue')" instead of ` +
              `"defineAsyncComponent(() => import('./MyPage.vue'))".`
```

src/navigationGuards.ts
```diff
@@ -294,6 +301,10 @@ export function extractComponentsGuards(
             const resolvedComponent = isESModule(resolved)
               ? resolved.default
               : resolved
+            // __asyncLoader is added by defineAsyncCompoent
+            if (__DEV__ && '__asyncLoader' in resolvedComponent) {
```
posva, line 305: Is this necessary? What does it change?

src/navigationGuards.ts
```diff
@@ -253,6 +253,13 @@ export function extractComponentsGuards(
           )
           let promise = rawComponent
           rawComponent = () => promise
+        } else if ((rawComponent as any)['__asyncLoader']) {
```
posva, line 256:
```suggestion
        } else if ((rawComponent as any).__asyncLoader) {
```

posva (changes requested):

posva (changes requested):
I just kept the same unresolved points!

leemove (commented):
> I just kept the same unresolved points!

Thank you for reviewing !

Those unresolved potints are all about the lazy load async component.
It's hard to distinguish between async component  and dynamic imports async component after cached.
Because  [the the following  code ](https://github.com/vuejs/vue-router-next/blob/468e6797a08a7772151000256d29f58eac4f5eaf/src/navigationGuards.ts#L305) cache the asyncComponent.

```js
  componentPromise.then(resolved => {
      const resolvedComponent = isESModule(resolved)
        ? resolved.default
        : resolved
      // replace the function with the resolved component
      record.components[name] = resolvedComponent
  }
```
compont will be `async componet` instead of `Promise<async component>`  when navigate to this route next time .

So I marked  the resolved async componet. 
```js
  if (__DEV__ && '__asyncLoader' in resolvedComponent) {
      resolvedComponent.__resolvedAsyncComponent = true
  }
```
and check the mark when async componet.
```js
 else if (
      (rawComponent as any).__asyncLoader &&
      !(rawComponent as any).__resolvedAsyncComponent // check  dynamic imports
    ) {
     // make it component promise ,  it will marked by code above, and warn only once when navigated first time.
      let asyncComponent = rawComponent
      rawComponent = () => Promise.resolve(asyncComponent)
      warn(.... )
    }
}
```
the next time navigate to this component, we can know it is  dynamic imports or not.

Sorry about my poor English. hope you  understand what I said.🙏

src/navigationGuards.ts
```diff
@@ -253,6 +253,18 @@ export function extractComponentsGuards(
           )
           let promise = rawComponent
           rawComponent = () => promise
+        } else if (
+          (rawComponent as any).__asyncLoader &&
+          !(rawComponent as any).__resolvedAsyncComponent
```
posva, line 258: let's remove the `__resolvedAsyncComponent` altogether and let the user fix the problem by themselves

src/navigationGuards.ts
posva: Is this accounting for `() => defineAsyncComponent(...)`? If it is, it should warn instead and it should also have a test

posva (changes requested):
Thanks! I think I understand what you said. I realized I wrote one of my comments at the wrong line. Let me know if the review isn't clear

src/navigationGuards.ts
leemove: It is also  accounting for `() => import(() => defineAsyncComponet())`

leemove: It is also accounting for 
```js
// MyPage.vue
export default defineAsyncComponent(...)
//router.js
() => import('./MyPage.vue')
```

leemove:
I remove the accounting for `() => defineAsyncComponent(...)` .

because it will warn `function that does not return a Promise`

and remove all warn for  function return Promise<AsyncComponent>,like `() => Promise<definAsyncComponent(...)>`

it will warn when navigate to this component next time .Beacuse next time,the component here is just `AsyncComponent`  not `Promise<AsyncComponent>`

posva (approved):
Thank you! I couldn't push some final changes (you should check the allow edits by maintainers) so I will do that locally

leemove:
Thank you for your patient review! I'm really excited about this PR was merged!

PR #714, opened 2021-01-12
docs: typo

Missmisery:
I think it's wrong here ：router.push('/admin').push(failure);
router.push return promise instance, push should be replaced with then.


posva (approved):
Nice catch, thanks!

PR #729, opened 2021-01-20
docs: improve guards

zhufengzhufeng:
Just return


docs/guide/advanced/navigation-guards.md
```diff
@@ -37,7 +37,7 @@ All of the the things above **work the same way with `async` functions** and Pro
 ```js
 router.beforeEach(async (to, from) => {
   // canUserAccess() returns `true` or `false`
-  return await canUserAccess(to)
+  return canUserAccess(to)
```
posva, line 40: Let's keep the await for the demo with something like
```suggestion
  const canAccess = await canUserAccess(to)
  if (!canAccess) return '/login'
```

posva (changes requested):

posva (approved):
Thanks!

PR #742, opened 2021-01-31
fix(devtools): id and label

Akryum:
Fixing the id (currently is `0`) and the label (more useful for user).


Akryum:
For context:
image

posva (approved):
Thanks!

PR #743, opened 2021-02-01
fix(guard): guard function not called for vue-class-component

### LINKED ISSUE #739: BeforeRouteEnter event not called for async loaded components
### Version
4.0.3

### Reproduction link
[https://github.com/websmurf/vue-router-async-test](https://github.com/websmurf/vue-router-async-test)

### Steps to reproduce
Run the provided application in the test repository, and make sure your console is open.

If you click the "Page 1" link, the console shows "Page 1"
If you click the "Page 2" link, the console shows "Page 2"
If you click the "Async Page 1" link, the console shows nothing
If you click the "Async Page 2" link, the console shows nothing

Now, click the Async Page 1 link again, now the console shows 'Async page 2'
Also, when directly visiting the async pages, no beforeRouteEnter event is triggered.

### What is expected?
I would expect the beforeRouteEnter method to be called in every of these cases.

The beforeRouteEnter method is only called for the sync loaded components. Only once the page is loaded once, the beforeRouteEnter method is called

---
We are in the process of migration our old Vue2 application to Vue3 and Vite, hence the usage of vue-class-component (since we use it already and it means converting lesser code).

Not sure if this is actually caused by vue-router; or by vue-class-component; will adjust the example to include a Composition API set as well.

ygj6:
close #739


src/navigationGuards.ts
```diff
@@ -316,8 +316,11 @@ export function extractComponentsGuards(
               : resolved
             // replace the function with the resolved component
             record.components[name] = resolvedComponent
+            // __vccOpts is added by vue-class-component and contain the regular options
+            let options: ComponentOptions =
+              (resolvedComponent as any).__vccOpts || resolvedComponent
             // @ts-ignore: the options types are not propagated to Component
```
posva, line 322: this should be removed now

posva (changes requested):
Thanks!

src/navigationGuards.ts
ygj6: I have removed the @ts-ignore , is that right?

posva (changes requested):
Can you rebase or remove all unrelated changes please?

PR #755, opened 2021-02-04
fix(guards): vue-class-component call guards on first...

### LINKED ISSUE #739: BeforeRouteEnter event not called for async loaded components

ygj6:
close #739

### DIFF
--- a/src/navigationGuards.ts
+++ b/src/navigationGuards.ts
@@ -316,8 +316,10 @@ export function extractComponentsGuards(
               : resolved
             // replace the function with the resolved component
             record.components[name] = resolvedComponent
-            // @ts-ignore: the options types are not propagated to Component
-            const guard: NavigationGuard = resolvedComponent[guardType]
+            // __vccOpts is added by vue-class-component and contain the regular options
+            let options: ComponentOptions =
+              (resolvedComponent as any).__vccOpts || resolvedComponent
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

websmurf:
Thanks!

PR #768, opened 2021-02-11
docs: typo

hootlex:
Fixes a small typo in the explanation:

> In this specific scenario we are using a [custom regexp](/guide/essentials/route-matching-syntax.md#custom-regexp-in-params) between parentheses and marking the `pathMatch` param as [optionally repeatable](/guide/essentials/route-matching-syntax.md#optional-parameters). This ~is to~ allows us to directly navigate to the route if we need to by splitting the `path` into an array:


posva (approved):
Thanks!

PR #769, opened 2021-02-11
docs: improve 404 example

hootlex:
The [code example in `pathMatch`](https://next.router.vuejs.org/guide/essentials/dynamic-matching.html#catch-all-404-not-found-route) results in the target URL starting with 2 slashes and gives the following error in the console.

```
this.$router.push({
  name: 'NotFound',
  params: { pathMatch: this.$route.path.split('/') },
})
```

image

I suggest to remove the first slash from the `path` using `substring(1)` before breaking it to an array. 

```
this.$router.push({
  name: 'NotFound',
  params: { pathMatch: this.$route.path.substring(1).split('/') },
})
```

The issue can be reproduced here: https://jsfiddle.net/j4pqmvh8/

I've also included a short notice in the docs.

### DIFF
--- a/docs/guide/essentials/dynamic-matching.md
+++ b/docs/guide/essentials/dynamic-matching.md
@@ -93,9 +93,10 @@ In this specific scenario we are using a [custom regexp](/guide/essentials/route
 ```js
 this.$router.push({
   name: 'NotFound',
-  params: { pathMatch: this.$route.path.split('/') },
+  params: { pathMatch: this.$route.path.substring(1).split('/') },
 })
 ```
+Notice that we should remove the first slash (`/`) from the path to avoid the target URL starting with with 2 slashes.
 
 See more in the [repeated params](/guide/essentials/route-matching-syntax.md#repeatable-params) section.
 
--- a/docs/guide/essentials/dynamic-matching.md
+++ b/docs/guide/essentials/dynamic-matching.md
@@ -93,13 +93,13 @@ In this specific scenario we are using a [custom regexp](/guide/essentials/route
 ```js
 this.$router.push({
   name: 'NotFound',
+  // preserve current path and remove the first char to avoid the target URL starting with `//`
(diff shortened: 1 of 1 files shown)


docs/guide/essentials/dynamic-matching.md
```diff
@@ -93,9 +93,10 @@ In this specific scenario we are using a [custom regexp](/guide/essentials/route
 ```js
 this.$router.push({
   name: 'NotFound',
-  params: { pathMatch: this.$route.path.split('/') },
+  params: { pathMatch: this.$route.path.substring(1).split('/') },
 })
 ```
+Notice that we should remove the first slash (`/`) from the path to avoid the target URL starting with 2 slashes.
```
posva, line 99: can you move this up as a comment?

docs/guide/essentials/dynamic-matching.md
```diff
@@ -93,9 +93,10 @@ In this specific scenario we are using a [custom regexp](/guide/essentials/route
 ```js
 this.$router.push({
   name: 'NotFound',
-  params: { pathMatch: this.$route.path.split('/') },
+  params: { pathMatch: this.$route.path.substring(1).split('/') },
```
posva, line 96:
```suggestion
  params: { pathMatch: this.$route.path.substring(1).split('/') },
  // preserve existing query and hash if any
  query: this.$route.query,
  hash: this.$route.hash,
```

posva (changes requested):
Thank you for this! I'm realizing we should also add the current query an hash when pushing

docs/guide/essentials/dynamic-matching.md
hootlex: great idea! It seemed a bit off to me here too.

hootlex:
@posva thanks for the prompt feedback. Changes applied 👍

I've also updated the wording of the notice to shorten it and describe what the line does. What do you think?

posva (approved):
Nice, thanks!

PR #777, opened 2021-02-17
refactor: use console.error during lazy loading

skyrpex:
By using `console.error` even in the development environment, we will provide useful information such as the reference to the actual ES module that failed and the code line where it happened.

This is especially important if the errors happen in an imported package, no matter how deep it is.

See the comparison here:

- With the current implementation, which uses `warn` in dev and `console.error` in prod:

image

You can see that the stack trace is of no help at all.

- With `console.error` after the usual `warn`:

image

Now, the browser points me to the module that failed.

By the way, let me please celebrate the [almost All Lucky 7s](https://finalfantasy.fandom.com/wiki/All_Lucky_7s) PR 😆

### DIFF
--- a/src/navigationGuards.ts
+++ b/src/navigationGuards.ts
@@ -298,9 +298,7 @@ export function extractComponentsGuards(
           componentPromise = Promise.resolve(componentPromise as RouteComponent)
         } else {
           // display the error if any
-          componentPromise = componentPromise.catch(
-            __DEV__ ? err => err && warn(err) : console.error
-          )
+          componentPromise = componentPromise.catch(console.error)
         }
 
         guards.push(() =>


src/navigationGuards.ts
```diff
@@ -298,9 +298,13 @@ export function extractComponentsGuards(
...
           // display the error if any
-          componentPromise = componentPromise.catch(
-            __DEV__ ? err => err && warn(err) : console.error
-          )
+          componentPromise = componentPromise.catch(err => {
+            if (__DEV__ && err) {
+              warn(err)
+            }
+
+            console.error(err)
```
posva, line 306: let's just keep the `console.error` if the warn is useless in dev

posva (changes requested):
Thanks for this! I don't think we will ever reach the lucky 7777 😆

src/navigationGuards.ts
skyrpex: Sounds good to me

skyrpex: Done.

posva (approved):
Thanks!

PR #780, opened 2021-02-19
Update index.md

PHPGuus:
It would be handy in this document to specify exactly how to get access to `VueRouter.createRouter`.


posva:
Thank you but this assumes plain index.html usage

PR #826, opened 2021-03-15
feat: terser support safari10

xue1206:
terser support safari10 to fix [a safari bug](https://stackoverflow.com/questions/42450473/safari-syntaxerror-cannot-declare-a-let-variable-twice)

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -195,6 +195,9 @@ function createMinifiedConfig(format) {
           ecma: 2015,
           pure_getters: true,
         },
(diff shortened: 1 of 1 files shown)


posva:
Can you check the allow maintainers edit so we can see the if there is a difference in the final bundle? I remember this not being necessary for vue router

posva:
Thank you but this doesn't seem to change the final bundle of vue router. Please open an issue with a reproduction if this is still failing

PR #844, opened 2021-03-24
feat(warn): throws if history is missing

### LINKED ISSUE #843: Not setting history option in initial config leads to confusing error
I'm trying to build my first Vue 3 project, and i'd like to include the new router as well. One thing i noticed is that you get an error that's a bit confusing when not providing the `history` option in the initial configuration, e.g.:

```
const router = createRouter({
    routes : [{
        path : '/',
        component : ScreenHome
    }]
});
```

Leads to an error:

`Uncaught TypeError: Cannot read property 'location' of undefined`

In the console. When i added the `createWebHashHistory` option everything worked fine. 

I know this is mentioned in the docs, but it's easy to miss, and coming from the Vue2 router (where history is set to hash by default) i think this will probably trip up many people trying to convert existing projects or starting new ones.

I think there would be three possible solutions:
1. Set a default `history` mode (like in v3)
2. Make the `history` option required
3. Provide an error message that's clearer ('Please provide a history option')

I would be really happy with either 1) or 2), but at least a better error message would be very helpful.

cheng-github:
Fix #843

### DIFF
--- a/__tests__/router.spec.ts
+++ b/__tests__/router.spec.ts
@@ -93,6 +93,11 @@ describe('Router', () => {
     createDom()
   })
 
+  it('fails if history option is missing', () => {
+    // @ts-ignore
+    expect(() => createRouter({ routes })).toThrow()
+  })
+
   it('starts at START_LOCATION', () => {
     const history = createMemoryHistory()
     const router = createRouter({ history, routes })
--- a/__tests__/router.spec.ts
+++ b/__tests__/router.spec.ts
@@ -95,7 +95,7 @@ describe('Router', () => {
 
   it('fails if history option is missing', () => {
     // @ts-ignore
-    expect(() => createRouter({ routes })).toThrow()
+    expect(() => createRouter({ routes })).toThrowError('Provide the "history" option')
   })
 
   it('starts at START_LOCATION', () => {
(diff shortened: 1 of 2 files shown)


src/router.ts
```diff
@@ -343,6 +343,11 @@ export function createRouter(options: RouterOptions): Router {
   let parseQuery = options.parseQuery || originalParseQuery
   let stringifyQuery = options.stringifyQuery || originalStringifyQuery
   let routerHistory = options.history
+  if (__DEV__ && !routerHistory)
+    throw new Error(
+      'Missing required history option of RouterOptions.\n' +
+        'See more at https://next.router.vuejs.org/api/#createrouter'
```
posva, line 349:
```suggestion
      'Provide the "history" option when calling "createRouter()":' +
        ' https://next.router.vuejs.org/api/#history.'
```

__tests__/router.spec.ts
```diff
@@ -93,6 +93,11 @@ describe('Router', () => {
     createDom()
   })
 
+  it('fails if history option is missing', () => {
+    // @ts-ignore
+    expect(() => createRouter({ routes })).toThrow()
```
posva, line 98:
```suggestion
    expect(() => createRouter({ routes })).toThrowError('Provide the "history" option')
```

posva (changes requested):
Thanks!

hay:
Tested and works as expected. Cheers.

PR #847, opened 2021-03-26
docs: use correct imports in code example

SpaceOtto:
The code example didn't work since it suggested that vue and vue-router have a named export which is not the case. To make it easier for beginners, I've added the correct imports to the code example.


posva:
Thank you for this but this is not exactly what we are going to do. More info at https://github.com/vuejs/vue-router-next/issues/829

laodengjuntuan:
thank you for your docs, it has solved my trouble.I can't figure out why official documentation has no the section about teaching us how to import  vue-router except cdn.

PR #854, opened 2021-03-29
docs: add description of key property when using transition in reused component

### LINKED ISSUE #803: Route with same component not triggering transition
### Version
4.0.4

### Reproduction link
[https://jsfiddle.net/mika76/1g2xn7j3/](https://jsfiddle.net/mika76/1g2xn7j3/)

### Steps to reproduce
Using transitions with routes as per https://next.router.vuejs.org/guide/advanced/transitions.html#transitions but transitions do not kick in if the component is the same. I've tried putting a :key on the router-view but it has not helped. Maybe some guidance on how this could be done in the docs?

### What is expected?
Transition should kick in on any route change

transition does not kick in if the component is the same (even if path is different)

shadowings-zy:
Close #803

### DIFF
--- a/docs/guide/advanced/transitions.md
+++ b/docs/guide/advanced/transitions.md
@@ -63,5 +63,17 @@ router.afterEach((to, from) => {
 })
 ```
 
+## Trigger Transition In Reused Component
+
+We need add `key` as a property of component to trigger `transition` correctly in reused component (multiple route path using the same component).
+
+```js
+<router-view v-slot="{ Component }">
+  <transition name="fade">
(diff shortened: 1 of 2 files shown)


posva:
Thanks! The idea is to show when adding a `key` can help. Most of the time people **shouldn't add a `key`**

shadowings-zy:
@posva hi, I update the PR. And I add the description "Trigger Transition In Reused Component"

the-adonis (approved):

posva:
Thank you for the PR but I went with a different version of this

PR #867, opened 2021-04-07
docs: add reuse-Component transition

### LINKED ISSUE #803: Route with same component not triggering transition

PeterAlfredLee:
Add reuse-Component transition doc.
Close #803

### DIFF
--- a/docs/guide/advanced/transitions.md
+++ b/docs/guide/advanced/transitions.md
@@ -63,5 +63,19 @@ router.afterEach((to, from) => {
 })
 ```
 
+## Reuse-Component Transition
+
+If there are multiple routing addresses using the same component,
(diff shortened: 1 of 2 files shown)


posva:
Thank you for the PR but I went with a different version of this

PR #872, opened 2021-04-08
feat: support for typescript named routes

### LINKED ISSUE #1160: Add support for params generic in useRoute
**Typing of params from useRoute.**

When you have function with typed arguments

```
const someFn = (id: string, type: ResourceType) => {}
```

And when you try to pass a `route.params` as an argument

```
const route = useRoute()
someFn(route.params.resourceId, route.params.type)
```
You will get error

Argument of type 'string | string[]' is not assignable to parameter of type 'string'.
  Type 'string[]' is not assignable to type 'string'.
  
Screenshot of problem

image

```
type Params = {
    id: string
    type: ResourceType
}
const route = useRoute<Params>() // route.params will be typed
```

### Pull Request
https://github.com/vuejs/vue-router-next/pull/1159

pikax:
WIP to support typescript named routes

image

~~## Limitations
`ExtractNamedRoutes` needs all the routes to have `children:[]` declare if one has `children`.
All the routes must have `name`~~

Close #1160

### DIFF
--- a/__tests__/errors.spec.ts
+++ b/__tests__/errors.spec.ts
@@ -16,7 +16,7 @@ import {
   RouteLocationNormalized,
 } from '../src/types'
 
-const routes: RouteRecordRaw[] = [
+const routes: Readonly<RouteRecordRaw>[] = [
   { path: '/', component: components.Home },
   { path: '/redirect', redirect: '/' },
   { path: '/foo', component: components.Foo, name: 'Foo' },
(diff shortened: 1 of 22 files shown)


pikax:
Should we also get the routes from the router?

image

The issue would be: 
- We only have access to the routes with that specific instance
- We would need to declare/override the default `"RouterOptions"` 
- How will it integrate with `NamedLocationMap`, should we merge both? I see `NamedLocationMap` to be used as the helper to define routes at runtime.
- How can we declare the `params` type?

sapphi-red:
https://vue-routider.sapphi.red/

I made a wrapper library which strictly types vue-router before.
I am not sure whether it is worth reading but I hope this helps you.

PR #879, opened 2021-04-13
docs: description of the use of hash and query together

PeterAlfredLee:


posva:
Thank you but I don't think this needs to be specified

PR #880, opened 2021-04-13
fix: do not allow invalid hazardous keys in query

cawa-93:
I drew attention to that when parsing the search query, it is possible to some extent manipulate the prototype of the returned object. I added a few checks so that at least partially prevent it.

### DIFF
--- a/src/query.ts
+++ b/src/query.ts
@@ -37,8 +37,7 @@ export type LocationQueryRaw = Record<
  * Do not allow invalid hazardous query keys
  */
 function isAllowedQueryKey(key: string): boolean {
-  const keysBlockList = ['__proto__', 'constructor', 'prototype']
-  return !keysBlockList.includes(key) && !Object.prototype.hasOwnProperty(key)
+  return !Object.prototype.hasOwnProperty(key)
 }
 
 /**
--- a/src/query.ts
+++ b/src/query.ts
@@ -34,13 +34,6 @@ export type LocationQueryRaw = Record<
 >
 
 /**
- * Do not allow invalid hazardous query keys
- */
-function isAllowedQueryKey(key: string): boolean {
-  return !Object.prototype.hasOwnProperty(key)
-}
-
-/**
  * Transforms a queryString into a {@link LocationQuery} object. Accept both, a
  * version with the leading `?` and without Should work as URLSearchParams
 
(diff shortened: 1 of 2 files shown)


cawa-93:
As a more rigorous and reliable alternative: return the `Map` or `URLSearchParams` instead of the `object`. Of course, this us breaking changes.

src/query.ts
```diff
@@ -55,6 +63,9 @@ export function parseQuery(search: string): LocationQuery {
     // allow the = character
     let eqPos = searchParam.indexOf('=')
     let key = decode(eqPos < 0 ? searchParam : searchParam.slice(0, eqPos))
+    if (!isAllowedQueryKey(key)) {
```
posva, line 66: Wouldn't this suffice to avoid setting invalid keys

```suggestion
    if (Object.prototype.hasOwnProperty(key)) {
```

`prototype` is a valid key BTW

cawa-93: I have abstracted logic into a separate function for easier readability and scaling.

posva: I'm saying that `Object.prototype.hasOwnProperty()` is the only check needed

cawa-93: @posva I simplified condition. But I propose to keep it in a separate function `isAllowedQueryKey`.

posva: Thanks, I think the extra function here adds no clarity but a little comment above saying this ignores `?__proto__&toString` is enough

posva (changes requested):
Thanks, I think this makes sense as it could avoid weird forced bugs when sharing a URL with intentionally wrong query params

posva (approved):
Thanks!

posva:
I think I will revert this change as it forbids keys like `toString` which are totally valid keys in objects and one can still avoid any set of query keys by provide a parseQuery and stringifyQuery.
`__proto__` cannot be overridden anyway.

Do you have any real scenario cases where this is a problem? Otherwise I will proceed to revert this change in the next patch @cawa-93

cawa-93:
No. I have no real scenarios. I see this as a safeguard against mistakes may by developer that lead to security risks

posva:
so what security risks where you thinking about? That's a scenario

cawa-93:
I thinking about Prototype pollution, but I don't have any real example. I only assume that there may be some scenarios that may have some scenarios may exist when the developer makes some non-standard behavior when it can lead to problems. I only **guess** that there may be some scenarios (maybe when the developer makes some non-standard behavior) when it can lead to problems.

posva:
Thanks for the input!

PR #883, opened 2021-04-14
docs: fix RouteMeta typing example

Bluenix2:
Without this change the example would have no exported members other than RouteMeta, a simple line of code such as:

```ts
import { createRouter } from 'vue-router'
```

Would throw the folloing error:

```
Module '"vue-router"' has no exported member 'createRouter'.
```


posva (approved):
Thanks! https://github.com/vuejs/vue-router-next/pull/790 opened the same but the tests I did where not correct and this is indeed necessary

Issue #855, opened 2021-03-30
Setting an optional parameter to undefined generates a link with `undefined` in the path instead of ommiting the paramater, like in vue-router@3


rashfael:
### Version
4.0.5

### Reproduction link
[https://jsfiddle.net/7sLubz0q/6/](https://jsfiddle.net/7sLubz0q/6/)

### Steps to reproduce
- click on link "Bar"

### What is expected?
url should show `/bar`, like in vue-router@3 https://jsfiddle.net/9bd0wag6/1/

### What is actually happening?
Url shows `/bar/undefined`.

posva:
You can only pass strings, numbers and arrays of these to params (https://github.com/vuejs/vue-router-next/blob/master/src/types/index.ts#L34). Vue Router 3 only accepts strings: https://github.com/vuejs/vue-router/blob/dev/types/router.d.ts#L193. This is why was never documented as a breaking change.

You can still achieve an empty optional param with an empty string or by omitting the key in the object

PR #885, opened 2021-04-15
Fix for #873 : replce router param value which is not a string or a number with empty string

### LINKED ISSUE #873: Document params must be strings, numbers, or arrays of these (for repeated params) in programmatic navigation
### Version
4.0.6

### Reproduction link
[https://codesandbox.io/s/dawn-pine-4yc98](https://codesandbox.io/s/dawn-pine-4yc98)

### Steps to reproduce
1. Press button on `Home` page which routes to `About` page with undefined `id` param.
2. Undefined `id` param is included in the URL & passed to `About` page as `string`

### What is expected?
Undefined params are not passed to URL & on the redirected page this param value should also be undefined

Undefined is passed to the URL & is usable on the destination page as a string

---
Also happens if `id` is passed as `null`.<br>
Doesn't matter if param is optional or not:<br>
* `path: "about/:id"`
* `path: "about/:id?`

PeterAlfredLee:
This is a fix #873 and #855.

There may be more similar issues in the future and I think we can fix this problem by simply replace the param values with empty strings.

### DIFF
--- a/__tests__/encoding.spec.ts
+++ b/__tests__/encoding.spec.ts
@@ -55,6 +55,21 @@ describe('Encoding', () => {
     it('encodes a specific charset', () => {
       expect(encodeParam(toEncode)).toBe(encodedToEncode)
     })
+
+    it('encodes undefined', () => {
+      // @ts-ignore
+      expect(encodeParam(undefined)).toBe('')
+    })
+
+    it('encodes false', () => {
+      // @ts-ignore
+      expect(encodeParam(false)).toBe('')
+    })
+
+    it('encodes true', () => {
+      // @ts-ignore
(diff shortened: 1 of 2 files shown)


posva:
Thank you but the issue is about documentation, not implementation. Also, keeping true and false as strings makes more sense than having an empty string

PeterAlfredLee:
> Also, keeping true and false as strings makes more sense than having an empty string

Yes, you are right. I didn't notice that before, thank you.

But what about `undefined`？I don't think people want to router to a `undefined` address. I think we can avoid this in our code. WDYT?

posva:
As said already, one should not pass undefined nor null

PeterAlfredLee:
ok

PR #912, opened 2021-04-28
fix($route): Make $route have an enumerable type

### LINKED ISSUE #902: Why the $route is not in the ctx but can be found and parsed in the template
### Version
4.0.6

### Reproduction link
[https://codesandbox.io/s/vue-router-route-73iju?file=/src/components/CustomComp.vue](https://codesandbox.io/s/vue-router-route-73iju?file=/src/components/CustomComp.vue)

### Steps to reproduce
1.
```html
<template>
  <div>
    <p>{{ $route }}</p>
  </div>
</template>
```

2.

```js
  setup() {
    const { ctx } = getCurrentInstance()
    onMounted(() => {
      console.log('Test $route: ', ctx.$route)
    })
    return {
      say: ref("Hello")
    };
  }
```

I know that `useRoute()` get `$route`, but I want to try use `ctx.$route`.
So I got `ctx` through `getCurrentInstance()` and then found that there is no `$route` field in `ctx` but it can be rendered normally in `<template>`.

I want to know why does it rendered properly in the `<template>` without the `$route` field in `ctx`?

I’m not sure if this is a bug or my usage is wrong. 

### What is expected?
console.log('Test $route: ', ctx.$route)

`ctx.$route` can be printed normally.

console.log('Test $route: ', ctx.$route)

`ctx.$route` is `undefined`

pdsuwwz:
Fix #902

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -1102,6 +1102,7 @@ export function createRouter(options: RouterOptions): Router {
 
       app.config.globalProperties.$router = router
       Object.defineProperty(app.config.globalProperties, '$route', {
+        enumerable: true,
         get: () => unref(currentRoute),
       })
 


posva (approved):
Thanks! This makes sense given how `$router` is defined. Note you should still not use `ctx` 😉

PR #924, opened 2021-05-06
Feat: support bashbang based url

### LINKED ISSUE #923: Base URL with a hashbang rewrites the URL breaking the routing
### Version
4.0.6

### Reproduction link
[https://codesandbox.io/s/keen-cannon-tmkfz](https://codesandbox.io/s/keen-cannon-tmkfz)

### Steps to reproduce
If you click on a link it works and renders the correct view and URL updates accordingly but if you can't refresh the URL as the createCurrentLocation method looks for a hash and places a forward slash "/" right after it which brakes the hashbang URL structure.

### What is expected?
If you refresh a route that has been rendered using routing the page should be able to be re-rendered based on the URL

The URL is being updated before it can be matched and so no route is found for the URL

---
While not a recommended best practice hashbang urls are still valid and should be rather trivial to support.  I specifically have an app that I am re-writing and in order to not break URLs need to support the hashbang structure

PeterAlfredLee:
Hashbang is not supported now. This PR provides an implemention about it.
See also [docs](https://developers.google.com/search/docs/ajax-crawling/docs/getting-started).

Fix #923

### DIFF
--- a/src/history/html5.ts
+++ b/src/history/html5.ts
@@ -39,11 +39,11 @@ function createCurrentLocation(
   location: Location
 ): HistoryLocation {
   const { pathname, search, hash } = location
-  // allows hash based url
-  const hashPos = base.indexOf('#')
-  if (hashPos > -1) {
-    // prepend the starting slash to hash so the url starts with /#
-    let pathFromHash = hash.slice(1)
+  // allows hash based or hashbang based url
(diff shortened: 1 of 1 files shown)


posva:
Thanks, I'm closing in favour of https://github.com/vuejs/vue-router-next/pull/925 which handles other cases too

PR #925, opened 2021-05-06
fix(hash): allow characters after `#` in base e.g. `#!`

### LINKED ISSUE #923: Base URL with a hashbang rewrites the URL breaking the routing

GGwujun:
Fix #923

### DIFF
--- a/src/history/html5.ts
+++ b/src/history/html5.ts
@@ -43,11 +43,11 @@ function createCurrentLocation(
   const hashPos = base.indexOf('#')
   if (hashPos > -1) {
     let slicePos =
-      hash.indexOf(base.slice(hashPos).toLowerCase()) > -1
+      hash.includes(base.slice(hashPos))
         ? base.slice(hashPos).length
         : 1
     let pathFromHash = hash.slice(slicePos)
-
+// prepend the starting slash to hash so the url starts with /#
     if (pathFromHash[0] !== '/') pathFromHash = '/' + pathFromHash
     return stripBase(pathFromHash, '')
   }
--- a/src/history/html5.ts
+++ b/src/history/html5.ts
@@ -39,11 +39,15 @@ function createCurrentLocation(
   location: Location
 ): HistoryLocation {
   const { pathname, search, hash } = location
-  // allows hash based url
+  // allows hash based or hashbang based url or others
   const hashPos = base.indexOf('#')
   if (hashPos > -1) {
-    // prepend the starting slash to hash so the url starts with /#
-    let pathFromHash = hash.slice(1)
+    let slicePos =
+      hash.indexOf(base.slice(hashPos).toLowerCase()) > -1
+        ? base.slice(hashPos).length
+        : 1
+    let pathFromHash = hash.slice(slicePos)
+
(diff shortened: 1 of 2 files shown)


edison1105:
duplicate of this PR: #924

GGwujun:
See also [docs](https://developers.google.com/search/docs/ajax-crawling/docs/getting-started).

Fix #923

posva (changes requested):
Thanks! Could you add a unit test to handle these bases

- `#!`
- `#other`

src/history/html5.ts
```diff
@@ -39,11 +39,15 @@ function createCurrentLocation(
...
 ): HistoryLocation {
   const { pathname, search, hash } = location
-  // allows hash based url
+  // allows hash based or hashbang based url or others
   const hashPos = base.indexOf('#')
   if (hashPos > -1) {
-    // prepend the starting slash to hash so the url starts with /#
-    let pathFromHash = hash.slice(1)
+    let slicePos =
+      hash.indexOf(base.slice(hashPos).toLowerCase()) > -1
```
posva, line 46:
```suggestion
      hash.includes(base.slice(hashPos))
```

src/history/html5.ts
```diff
@@ -39,11 +39,15 @@ function createCurrentLocation(
   location: Location
 ): HistoryLocation {
   const { pathname, search, hash } = location
-  // allows hash based url
+  // allows hash based or hashbang based url or others
   const hashPos = base.indexOf('#')
   if (hashPos > -1) {
-    // prepend the starting slash to hash so the url starts with /#
```
posva, line 45: can you add back the comment at line 50

posva (changes requested):
Thanks!

posva:
I added tests that were failing before. They seem to be failing when ran with other tests but work when ran alone. It will need more investigation

src/history/html5.ts
```diff
@@ -39,11 +39,15 @@ function createCurrentLocation(
...
-    let pathFromHash = hash.slice(1)
+    let slicePos =
+      hash.indexOf(base.slice(hashPos).toLowerCase()) > -1
+        ? base.slice(hashPos).length
+        : 1
+    let pathFromHash = hash.slice(slicePos)
+
```
posva, line 50:
```suggestion
// prepend the starting slash to hash so the url starts with /#
```

src/history/html5.ts
GGwujun: good,thks

posva (approved):

PR #927, opened 2021-05-07
fix(hash): second time redirect page no effect

### LINKED ISSUE #916: Redirect the page through the router guard, the second time will not take effect
### Version
4.0.6

### Reproduction link
[https://github.com/XingXiaoWu/ele-error](https://github.com/XingXiaoWu/ele-error)

### Steps to reproduce
# Note: 
I have 2 path, `/about` and `/login`, and if url don't inclued `login`, I will redirect to `/login`
# code
```
import { createRouter, createWebHashHistory } from 'vue-router'
import Home from '../views/Home.vue'

const routes = [
  {
    path: '/about',
    name: 'About',
    // route level code-splitting
    // this generates a separate chunk (about.[hash].js) for this route
    // which is lazy-loaded when the route is visited.
    component: () => import(/* webpackChunkName: "about" */ '../views/About.vue')
  },
  {
    path: '/login',
    name: 'Login',
    component: Home,
  }
]

const router = createRouter({
  history: createWebHashHistory(),
  routes
})

router.beforeEach((to)=>{
  debugger
  if (to.path.includes('/login')) {
    return true;
  }else{
    return {
      path:'/login'
    }
  }
})

export default router

```
# Step:
1. Type `http://localhost:8081/#/about` in your browser's address bar

it will redirect to `http://localhost:8081/#/login`

2. Type `http://localhost:8081/#/about`  again

### What is expected?
the page is login, 
and the browser's address bar is `http://localhost:8081/#/login`

the page is login
But the browser's address bar is `http://localhost:8081/#/about`

---
Pages also change on some micro front-end frameworks, like [qiankun](https://github.com/umijs/qiankun)

PeterAlfredLee:
This PR is about #916.

I think we should not throw `NAVIGATION_DUPLICATED` if `redirectedFrom` exists, and at the same time `from`, `targetLocation` & `redirectedFrom` are not the same value.

I tested this PR with the case in #916 and it works. Feel free to let me know if I need to add the corresponding tests. :-)

Fix #916

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -638,22 +638,29 @@ export function createRouter(options: RouterOptions): Router {
     toLocation.redirectedFrom = redirectedFrom
     let failure: NavigationFailure | void | undefined
 
-    if (!force && isSameRouteLocation(stringifyQuery, from, targetLocation)) {
-      failure = createRouterError<NavigationFailure>(
-        ErrorTypes.NAVIGATION_DUPLICATED,
-        { to: toLocation, from }
-      )
-      // trigger scroll to allow scrolling to the same anchor
-      handleScroll(
-        from,
-        from,
-        // this is a push, the only way for it to be triggered from a
-        // history.listen is with a redirect, which makes it become a push
-        true,
-        // This cannot be the first navigation because the initial location
-        // cannot be manually navigated to
-        false
-      )
+    if (!force) {
+      let isSame = isSameRouteLocation(stringifyQuery, from, targetLocation)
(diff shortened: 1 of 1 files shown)


PeterAlfredLee:
> Thank you but I'm not sure about this. The correct way would be to restore the history like in #921. Did you find a case where that fix was failing?

Hi @posva . I think I find a solution about this.
I pushed a new PR #944 . You can merge #944 if you think it's a good solution, or you can close that PR and modify the code of this PR. It's up to you. :)

PR #935, opened 2021-05-09
fix(query): `&` sign doesn't remove properly

### LINKED ISSUE #934: Redundant ampersand when trying to remove query param
### Version
4.0.6

### Reproduction link
[https://codesandbox.io/s/priceless-https-bvip5](https://codesandbox.io/s/priceless-https-bvip5)

### Steps to reproduce
1. Open reproduction link, take a look at the url and just wait for 2 timeouts.
2. When second timeout is passed you'll notice unnecessary ampersand sign at the end of the current url.

### What is expected?
When I pass `undefined` or empty array as a query parameter value then I expect query parameter will be **completely** removed from the current url.

Query parameter was properly removed but unnecessary `&` sign still presents at the end of the url.

---
I believe that's because of this line: https://github.com/vuejs/vue-router-next/blob/ecd52e017ac30fa996d4796974371374f65640d1/src/query.ts#L92
It doesn't take into account the passed value and appends `&` sign right away no matter whether the value is null or undefined.

mattersj:
Fix #934

### DIFF
--- a/__tests__/stringifyQuery.spec.ts
+++ b/__tests__/stringifyQuery.spec.ts
@@ -47,4 +47,11 @@ describe('stringifyQuery', () => {
   it('keeps = in value', () => {
     expect(stringifyQuery({ a: '=' })).toEqual('a==')
   })
+
+  it('removes values properly', () => {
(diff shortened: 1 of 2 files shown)


posva:
Ended up keeping the behavior but created multiple tests. Thanks again

mattersj:
You are welcome!

PR #941, opened 2021-05-11
docs: use beforeResolve

DeligiannisGeorge:
I think beforeEach example should be inside beforeEach section

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -501,6 +501,14 @@ _Parameters_
 | --------- | ------------------------------------- | ----------------------- |
 | guard     | [`NavigationGuard`](#navigationguard) | navigation guard to add |
 
+#### Examples
+
+```js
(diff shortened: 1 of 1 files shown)


docs/api/index.md
```diff
@@ -517,14 +525,6 @@ _Parameters_
 | --------- | ------------------------------------- | ----------------------- |
 | guard     | [`NavigationGuard`](#navigationguard) | navigation guard to add |
 
-#### Examples
-
-```js
-router.beforeEach(to => {
```
posva, line 523: let's keep the example but use `beforeResolve`

posva (changes requested):

docs/api/index.md
DeligiannisGeorge: Done

posva (approved):

PR #944, opened 2021-05-14
Fix the lint check failure of #921

PeterAlfredLee:
This PR is a fix of #921 .

The declear of `changeURL` seems doesn't work and the lint check still reports.

I think we can make the `routerHistory` to be a `any` object like this :
```javascript
(routerHistory as any).changeURL = function (url: string) {
```

WDYT?


posva:
Thanks but no, I prefer using the ts-ignore or ts-error. Could you comment on the PR instead of making a PR to master, please?

posva:
Thanks for the heads up on the lint error, I fixed that with a different approach in the original PR

PR #949, opened 2021-05-17
docs: update history mode example

PeterAlfredLee:
1. Fix some typos.
2. Update `history-mode.md`, add an example that will work for Caddy v2.


posva (approved):
Thanks!

PR #974, opened 2021-05-31
docs: note about new props when using props: true

Soviut:
Highlighted the fact that in order for `props: true` to work, you need to explicitly define those props and name them correctly for the values to pass.

### DIFF
--- a/docs/guide/essentials/passing-props.md
+++ b/docs/guide/essentials/passing-props.md
@@ -15,6 +15,7 @@ with
 
 ```js
 const User = {
+  // prop names must match route params for values to be passed
   props: ['id'],
   template: '<div>User {{ id }}</div>'
 }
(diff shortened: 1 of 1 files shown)


posva (changes requested):
Let's show what user.vue should look like instead

Soviut:
@posva There is a code example directly above with a simple `User` component, but it's very easy to miss the additional `props: ['id']` added to it (I did initially). Since I was using the composition API I thought perhaps that the route.params were automatically added to `setup(props)` and was wondering why the passed `props` were empty.

posva:
Then let's add a comment above `props: [...]`

Soviut:
@posva I've moved the note into a comment in the code example.

docs/guide/essentials/passing-props.md
```diff
@@ -15,6 +15,7 @@ with
 
 ```js
 const User = {
+  // prop names must match route params for values to be passed
```
posva, line 18:
```suggestion
  // make sure to add a prop named exactly like the route param
```

posva (approved):
Thanks!

PR #976, opened 2021-05-31
docs: fix vertical bar in tables

fz6m:


docs/.vitepress/theme/index.js
```diff
@@ -1,5 +1,6 @@
 import DefaultTheme from 'vitepress/dist/client/theme-default'
 import Layout from './Layout.vue'
+import './styles/index.css'
```
posva, line 3: I pushed a change in CSS for this, you can remove it. It will avoid adding `div`s

posva (changes requested):
Thank you! Can you rebase from master? I added the needed CSS to avoid having to use `div`s

posva (approved):
Thanks!

docs/.vitepress/theme/index.js
fz6m: Great. I deleted this css file

PR #977, opened 2021-06-02
docs: update the url of SSR

PeterAlfredLee:
Update the url link in documents written in Chinese.


posva (changes requested):
Let's use the v3 links:

- https://v3.vuejs.org/guide/ssr/introduction.html#what-is-server-side-rendering-ssr

It looks like the Chinese docs are still missing that page

posva (approved):
Thanks!

PR #990, opened 2021-06-08
chore(playground): unused '@ts-expect-error' directive

PeterAlfredLee:
There is a error when run command `yarn dev`:
```
./playground/main.ts 13:4-23
[tsl] ERROR in D:\github\vuejs\vue-router-next\playground\main.ts(13,5)
      TS2578: Unused '@ts-expect-error' directive.
```
I think we should remove the comment `// @ts-expect-error` in `playground/main.ts`.

### DIFF
--- a/playground/main.ts
+++ b/playground/main.ts
@@ -10,7 +10,6 @@ declare global {
     // h: HTML5History
     h: typeof routerHistory
     r: typeof router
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

PR #1002, opened 2021-06-16
feat(link): expose useLink on RouterLink as internal

KaelWD:
This allows us to make vue-router optional while still being able to use the composable instead of router-link itself:
```ts
import { resolveDynamicComponent } from 'vue'
import type {
  RouterLink as _RouterLink,
  useLink as _useLink,
  UseLinkOptions,
} from 'vue-router'

function useLink (props: Partial<UseLinkOptions>): ReturnType<typeof _useLink> | undefined {
  const RouterLink = resolveDynamicComponent('RouterLink') as typeof _RouterLink | string

  if (typeof RouterLink === 'string') return

  const link = RouterLink.useLink(props as UseLinkOptions)

  return props.to
    ? link
    : undefined
}
```

### DIFF
--- a/src/RouterLink.ts
+++ b/src/RouterLink.ts
@@ -149,6 +149,8 @@ export const RouterLinkImpl = /*#__PURE__*/ defineComponent({
     },
   },
 
+  useLink,
+
   setup(props, { slots }) {
     const link = reactive(useLink(props))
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

PR #1004, opened 2021-06-17
Update composition-api.md

ndianabasi:
Add the `next` NavigationGuard callback to enhance default behaviour of continuing navigation when return is `void` or `true` or `undefined`.

This is an appropriate behaviour in production.


posva:
Thank you but the  current version is the correct one

PR #1007, opened 2021-06-21
fix typescript error about Type undefined is not assignable to type object

heart-er:
fix typescript error about Type undefined is not assignable to type object


posva:
Thank you for the PR. There was a simpler check at bedd350 to fix the typing error

PR #1021, opened 2021-06-30
fix(guards): components guards in mixins

webigorkiev:
https://github.com/vuejs/vue-router-next/issues/454
Maybe not a good idea, but since you are using rawComponent it makes sense.
Probably it was worth on the vue side to create the ability to resolve component options, just as it exists for an instance.

As written here:
https://github.com/vuejs/vue-next/issues/2119


posva:
Thank you but this won't work with extended components nor global mixins

posva:
I really appreciate your work on this but it really needs the mentioned feature to work in all cases. I rather let you know that let you waste more time on this 🙃.

PR #1033, opened 2021-07-08
feat(devtools): add more options in devtools routes

PeterAlfredLee:
What problem does this feature solve?

This PR support displaying location of vue sfc, some meta data and whether or not the route is matched. These things will be displayed at right sidebar in Routing->Routes. I think these may be of help for developing.


posva:
Thanks, I tested locally and unfortunately most of these options do not work in common scenarios so I ended up keeping only a few of them on 051b7c7

PR #1037, opened 2021-07-11
Add extra instructions to import router lib

iamntz:
The documentaion is unclear and for a beginner is really confusing.


posva:
Thank you but those changes would be incorrect given the beginning of the docs. See https://github.com/vuejs/vue-router-next/issues/829

PR #1042, opened 2021-07-13
Update index.md

zdRan:
update example code


posva:
Thank you but this is correct. See open issues for docs

PR #1059, opened 2021-08-02
fix(types): Support `undefined` in NavigationGuardNext 

Thomasan1999:
Add support for `undefined` as the first argument in `next` function in navigation guards. I've decided to do it through making `location` argument optional since it seems to me as using `undefined` is omitting redefining the next route location.

Related to: https://github.com/vuejs/vue-router-next/issues/1058

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -367,7 +367,6 @@ export interface MatcherLocation
   > {}
 
 export interface NavigationGuardNext {
-  (): void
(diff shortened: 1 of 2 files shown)


posva (changes requested):
Let's add it to the Boolean argument instead as an explicit undefined. Can you also add a test in the dts tests?

posva (approved):
Thanks!

PR #1061, opened 2021-08-05
correcting path to params in route object

BullwinkleDev:


posva:
Thank you but this is correct. You are probably using `useRouter()` instead of `useRoute()`

BullwinkleDev:
Yes! Sorry my bad

чт, 5 авг. 2021 г., 12:47 Eduardo San Martin Morote <
***@***.***>:

PR #1070, opened 2021-08-12
fix(ssr): memory leak, only track installedApps in browser

tjk:
Could there be cases where some does want to mount + unmount on server side? An alternative could be to return the cleanup function so it can be called manually in environments were apps are created and disposed of? Seems like more of an edge case...

EDIT: or should we be unmounting once render is complete? that is not in docs or the examples... let me see about that.

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -1201,20 +1201,22 @@ export function createRouter(options: RouterOptions): Router {
       app.provide(routeLocationKey, reactive(reactiveRoute))
       app.provide(routerViewLocationKey, currentRoute)
 
-      const unmountApp = app.unmount
-      installedApps.add(app)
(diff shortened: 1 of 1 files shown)


posva:
Where is the leak exactly?
Usually, on the server, each request creates a whole new app with its router, etc and at then end of the request the the whole app is disposed

tjk:
Excuse me -- I misread this file. Been staring at memory profiles for way too many hours, see `installedApps` littered in a bunch of a retained object but it does look like it's scoped. Something else is causing app objects to not be reaped and I'm still searching. Sorry about that.

posva:
No worries, please let me know if you find more as the reference could still be held by the router.

PR #1076, opened 2021-08-16
docs: note about router version

devzakir:


posva:
Thank you but this version of the docs doesn't need the tip like v3

PR #1077, opened 2021-08-16
add in the way to use it

marcus7777:
this is where I got lost and confused


posva:
Thank you but this is not the way to use it. VueRouter is accessible because it's included as a global script.

---

Please, search existing issues and pull requests before opening issues, it saves maintainers a lot of time 🙏

PR #1101, opened 2021-09-02
fix: warn when multiple instances of router package are used in one app

znck:
Demo: https://github.com/znck/double-router-demo
Codesandbox: https://githubbox.com/znck/double-router-demo


posva:
I think this is too complicated for a warning. If it can be implemented with one single check (e.g. on the most commonly used feature like RouterView) then it could be added but also I haven't seen any reports of this issue so I don't such a niche case is worth adding

PR #1112, opened 2021-09-09
fix(history): ensure base is normalized in memory history

david50407:
Ensure base is normalized in memory history.

When using memory history with slash-trailing base (e.g. `/foo/`), `router.resolve().href` will combine base and normalized location (e.g. `/bar/`) directly, this will cause double-slashes in href url.
To fix this issue, ensure base is normalized just like we have done in html5/hash history.


posva (approved):
Nice catch, thanks!

PR #1121, opened 2021-09-17
[fix]:Optimize the quick start module used in vue3 (ENGLIST)

piexlmax:


posva:
Thank you but this is not correct

PR #1157, opened 2021-10-16
build: use cjs/mjs extensions for cjs/esm builds

danielroe:
## Summary

This PR proposes using `.cjs` and `.mjs` instead of the current `.js` extensions. That will allow usage of this library in a native Node module context.

| **old**                   | **new**                    |
| ------------------------- | -------------------------- |
| vue-router.cjs.js         | vue-router.cjs             |
| vue-router.cjs.prod.js    | vue-router.prod.cjs        |
| vue-router.esm-bundler.js | vue-router.mjs             |

## Notes

* As is, this is a **breaking change** as it changes the packaging expectations. I've added stubs that re-export these `.mjs`/`.cjs` files for backwards compatibility.
* Because this package imports `vue`, `vue` will also need to provide appropriately named native Node module filenames. (This should not negatively affect existing users who will not be using this module natively anyway.)

## Background issue

At the moment, importing `vue-router` within an ESM context produces the following error:

```
(node:73025) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
Server Side Rendering Error: /project/node_modules/vue-router/dist/vue-router.esm-bundler.js:6
import { getCurrentInstance, inject, onUnmounted, onDeactivated, onActivated, computed, unref, watchEffect, defineComponent, reactive, h, provide, ref, watch, shallowRef, nextTick } from 'vue';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:1001:16)
    at Module._compile (internal/modules/cjs/loader.js:1049:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:14)
    at ModuleWrap.<anonymous> (internal/modules/esm/translators.js:199:29)
    at ModuleJob.run (internal/modules/esm/module_job.js:169:25)
    at async Loader.import (internal/modules/esm/loader.js:177:24)
    at async __instantiateModule__ (file:///project/playground/.nuxt/dist/server/server.mjs:1844:3)
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -6,7 +6,7 @@
 
 <!--email_off-->
 
-[Unpkg.com](https://unpkg.com) provides npm-based CDN links. The above link will always point to the latest release on npm. You can also use a specific version/tag via URLs like `https://unpkg.com/vue-router@4.0.5/dist/vue-router.global.js`.
+[Unpkg.com](https://unpkg.com) provides npm-based CDN links. The above link will always point to the latest release on npm. You can also use a specific version/tag via URLs like `https://unpkg.com/vue-router@4.0.15/dist/vue-router.global.js`.
 
 <!--/email_off-->
 
--- a/docs/zh/installation.md
+++ b/docs/zh/installation.md
@@ -6,7 +6,7 @@
 
 <!--email_off-->
 
-[Unpkg.com](https://unpkg.com) 提供了基于 npm 的 CDN 链接。上述链接将始终指向 npm 上的最新版本。 你也可以通过像 `https://unpkg.com/vue-router@3.0.0/dist/vue-router.cjs` 这样的 URL 来使用特定的版本或 Tag。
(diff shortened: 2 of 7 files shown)


posva:
Published and testable [under the `egde` tag](https://www.npmjs.com/package/vue-router/v/4.0.0-329e962), `vue-router@edge`

src/globalExtensions.ts
```diff
@@ -62,4 +62,9 @@ declare module '@vue/runtime-core' {
     RouterView: typeof RouterView
     RouterLink: typeof RouterLink
   }
+
+  export interface GlobalComponents {
+    RouterView: typeof RouterView
+    RouterLink: typeof RouterLink
+  }
```
posva, line 69:
```suggestion
```

PR #1159, opened 2021-10-16
types(useRoute): add generic for params

fyapy:
### Version
4.0.12

### Steps to reproduce
Pass params to function with typed arguments

image

Error message
Argument of type 'string | string[]' is not assignable to parameter of type 'string'.
  Type 'string[]' is not assignable to type 'string'.

### What is expected?
Possability to pass generic for params types, like in react-router-dom hook useParams
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/09ce5bbfc2d507d35f6e8185a1f6a617bfca20fc/types/reach__router/index.d.ts#L182

### How will it look like
`type Params = {`
`    id: string`
`    type: ResourceType`
`}`
`const route = useRoute<Params>()`

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -105,7 +105,7 @@ export interface RouteLocationMatched extends RouteRecordNormalized {
  *
  * @internal
  */
-export interface _RouteLocationBase {
+export interface _RouteLocationBase<Params extends RouteParams = RouteParams> {
   /**
    * Percentage encoded pathname section of the URL.
    */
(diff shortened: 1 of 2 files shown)


posva:
Thank you for this. Let's not add new generics yet because then it won't be possible to adapt without a new major release and a breaking change 😉 .

For the moment, the best solution is to have a user-defined `useParams()` like mentioned at https://github.com/vuejs/router/issues/1160#issuecomment-956037512 that overrides the types for params. Something like:

```ts
import { RouteParams } from 'vue-router'
import { computed } from 'vue'

function useParams<P extends RouteParams>() {
  const route = useRoute()
  return computed(
    () => route.params as P
  )
}
```

1nusah:
@posva , hi any news on this? I'm interested in finalising this so it can be supported in nuxt3

PR #1163, opened 2021-10-19
Update index.md

adamkpurdy:
The current instructions seem backwards. Shouldn't you use the `custom` prop to trigger the custom element to wrap the inner element?


posva:
Thanks but the current docs are correct.

adamkpurdy:
Ok, I see. The example was slightly confusing to me with using an a tag when using the `custom` prop. Thank you for your quick response and your work on vue-router.

PR #1177, opened 2021-11-01
Fix qs example [ci skip]

machour:
Currently referencing the non-existant `parse` property instead of `parseQuery`


posva (approved):
Thanks!

PR #1186, opened 2021-11-03
feat(types): let property of `RouteParams` optional (fix #1184)

### LINKED ISSUE #1184: Property of type RouteParams and RouteParamsRaw should be optional
Enabling [`@typescript-eslint/no-unnecessary-condition`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md), then access `Number(route.params.example ?? 0)` will trigger
>ESLint: Unnecessary conditional, expected left-hand side of `??` operator to be possibly null or undefined.(@typescript-eslint/no-unnecessary-condition)

Switch to `Number(route.params.example)` might give me an NaN if current route doesn't have a param named `example`.

I know I can suppress this lint rather than make all property of `RouteParams` optional, which will bother other users since they will have to write undefined safe type guard when accessing params (like `route.params?.example`), but it's for runtime type safety since there's no guarantee for the existence of params value.

https://github.com/vuejs/vue-router-next/blob/22739c3918f1ce28e845a72019f85290bed9629d/src/types/index.ts#L35-L39
can be modify with
```ts
export type RouteParams = { [P in string]?: RouteParamValue | RouteParamValue[] };
export type RouteParamsRaw = { [P in string]?: RouteParamValueRaw | Array<Exclude<RouteParamValueRaw, null | undefined>> };
```

n0099:
BREAKING CHANGE: Users will have to write undefined type guards against `route.param`

close #1184

### DIFF
--- a/src/RouterLink.ts
+++ b/src/RouterLink.ts
@@ -302,8 +302,8 @@ function includesParams(
     } else {
       if (
         !Array.isArray(outerValue) ||
-        outerValue.length !== innerValue.length ||
-        innerValue.some((value, i) => value !== outerValue[i])
+        outerValue.length !== innerValue?.length ||
+        innerValue?.some((value, i) => value !== outerValue[i])
       )
         return false
     }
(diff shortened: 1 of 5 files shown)


posva:
Thanks for the PR. Having the breaking change is, unfortunately, going to be a no-go. We will instead go for something that enables 

- https://github.com/vuejs/router/issues/1160

PR #1190, opened 2021-11-04
feat(view): handle empty components as pass through

### LINKED ISSUE #1174: Allow omitting `component`/`components` in a route with `children`
This would allow avoiding creating empty `() => h(RouterView)` wrappers and losing any wrapping transition/suspense (https://github.com/vuejs/vue-router-next/issues/1173)

The type should allow this definition:

```js
{
  path: '/users'
  children: [
    { path: '', component: UserList },
    { path: ':id', component: UserDetails }
  ]
}
```

posva:
Closes #1174

### DIFF
--- a/__tests__/RouterView.spec.ts
+++ b/__tests__/RouterView.spec.ts
@@ -213,7 +213,6 @@ const routes = createRoutes({
     meta: {},
     matched: [
       {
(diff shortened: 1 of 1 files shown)


Thanayaby:
Mta>=

PR #1196, opened 2021-11-10
docs: better example

perenstrom-bontouch:
I found that I had to re-read this section a couple of times, to figure out which was `a` and which was `b`. It would be easier to understand if the text reflected the example.

### DIFF
--- a/docs/guide/essentials/redirect-and-alias.md
+++ b/docs/guide/essentials/redirect-and-alias.md
@@ -2,7 +2,7 @@
 
 ## Redirect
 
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

jsw3363:
@posva 

Excuse me. I'm so sorry if I can ask you a question here but, could I take contributing - contributing guide?

It's difficult because of the low priority, can I take docs? The language is Korean

jsw3363:
@posva 

Excuse me. I'm so sorry if I can ask you a question here but, could I take contribution - contributing guide?

It's difficult because of the low priority, can I take docs? The language is Korean

PR #1206, opened 2021-11-16
default routerHistory mode

zhangenming:


zhangenming:
So why not provide a default value is there any other consideration thank you( https://github.com/vuejs/vue-router-next/issues/843 option 1)

posva:
Thanks but this is intentional for tree shaking.

zemmmmmmm:
So why not
I like that

PR #1211, opened 2021-11-20
docs: add explicit import

Sumurai8:
The syntax used to be `import VueRouter from 'vue-router'`, but this does not seem to work anymore. With some experimentation this seems to be the required syntax, but I can't find the documentation for that anywhere. With this change it will be clear from the code snippet how to properly import the router.

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -44,6 +44,8 @@ Note how instead of using regular `a` tags, we use a custom component `router-li
 ## JavaScript
 
 ```js
+import * as VueRouter from 'vue-router'
+
 // 1. Define route components.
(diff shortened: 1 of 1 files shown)


posva:
This is actually intentional. There is an open issue about it

Sumurai8:
I am guessing you are referring to [this cryptic issue](https://github.com/vuejs/vue-router-next/issues/829) from 7 months ago. I fail to see how that issue detailing a full rewrite of the docs prevents us from having the short-term benefit of mentioning anywhere how to import the router. As I see it, you can do both, and it prevents people from bumping in yet another roadblock while trying out this version of the router while you improve the docs long-term.

PR #1214, opened 2021-11-21
In documentation, chain use(router) and mount()

alberto56:
Making changes to the data model (app.my_custom_variable = 'hello world') works when you chain app.use(router).mount('#app'), but may not reflect in the DOM when you don't chain them (app.use(router); app.mount('#app')).


posva:
Thanks but this syntax is easier for beginners

PR #1234, opened 2021-12-07
docs(zh): fix incorrect translation

zgayjjf:
It is `in the above example` in English doc, so it should be `上面的例子中` instead of `下面的例子中` in Chinese doc.


zgayjjf:
The origin sentence:
https://github.com/vuejs/vue-router-next/blob/master/docs/guide/essentials/redirect-and-alias.md?plain=1#L37

posva (approved):
Thanks! This looks good to me but just to make sure 下面 and 上面 do not have different usage in text, @Justineo can you help me confirm, please?

Justineo (approved):

PR #1267, opened 2022-01-14
feat: runtime warning for invalid `components` option

### LINKED ISSUE #1260: When defining named views get unknown error
### Version
4.0.12

### Reproduction link
[jsfiddle.net/6hnx3yfz/](https://jsfiddle.net/6hnx3yfz/)

### Steps to reproduce
When defining multiple named views, as you know you need to define like below:
```
            {
                path: '/foo',
                components: {default:Foo, Bar},
                name: 'foo',
            },
```

I mistakenly realized when you define component like down below:
```
            {
                path: '/foo',
                components: Foo,
                name: 'foo',
            },
```
When you define like this you will get some errors in  console and you can not change route, shortly your project is exploded. Actual point is console don't telling you what is wrong.

### What is expected?
Errors need to be explanatory. Example: "Route Components  need to be object".

You got some error but console don't telling you what is wrong. 

---
I was trying named views and i rolled back but forgot the change the "components" key in defined route object. After 1 hour I realized whats going on.

eddie0329:
## Reproduction Link

- [jsfiddle link](https://jsfiddle.net/6hnx3yfz/)

Thanks to @YusufcanY reproduce bug.

## Works

Add warns when route record with "components" property. Cover warning cases mentioned issue #1260.

- Warning and throws non object values( ex> components or functional components) are placed.
- Warning miss behaved lazy loading.  ( ex> import(./Mypage.vue) )

fix #1260

### DIFF
--- a/__tests__/matcher/addingRemoving.spec.ts
+++ b/__tests__/matcher/addingRemoving.spec.ts
@@ -1,6 +1,8 @@
-import { createRouterMatcher } from '../../src/matcher'
+import { createRouterMatcher, RouterMatcher } from '../../src/matcher'
 import { MatcherLocation } from '../../src/types'
 import { mockWarn } from 'jest-mock-warn'
+import { components } from '../utils'
+import { ErrorTypes } from '../../src/errors'
 
 const currentLocation = { path: '/' } as MatcherLocation
 // @ts-expect-error
@@ -410,6 +412,37 @@ describe('Matcher: adding and removing records', () => {
         {}
       )
       expect('same param named').not.toHaveBeenWarned()
+    })
+
+    describe(`not appropriate values placed in components property"`, () => {
+      let matcher: RouterMatcher
+      beforeEach(() => {
+        matcher = createRouterMatcher([], {})
+      })
+
+      it('not appropriate object placed in components propery -> warn + error', () => {
+        try {
+          matcher.addRoute({
+            path: '/',
+            components: components.Home,
+            name: 'home',
+          })
+        } catch ({ message }) {
+          expect(message).toBe('Invalid route record "components" property.')
+          expect('should be object.').toHaveBeenWarned()
+        }
+      })
+
+      it('not appropriate lazyload in components property -> warn', () => {
+        matcher.addRoute({
+          path: '/',
+          components: { default: import('../utils') },
+          name: 'home',
+        })
+        expect(
+          'Promise instead of a function that returns a Promise.'
+        ).toHaveBeenWarned()
+      })
     })
   })
 })
--- a/__tests__/matcher/addingRemoving.spec.ts
+++ b/__tests__/matcher/addingRemoving.spec.ts
@@ -1,7 +1,6 @@
 import { createRouterMatcher, RouterMatcher } from '../../src/matcher'
 import { MatcherLocation } from '../../src/types'
 import { mockWarn } from 'jest-mock-warn'
-import { components } from '../utils'
 
 const currentLocation = { path: '/' } as MatcherLocation
 // @ts-expect-error
--- a/__tests__/matcher/addingRemoving.spec.ts
+++ b/__tests__/matcher/addingRemoving.spec.ts
@@ -2,7 +2,6 @@ import { createRouterMatcher, RouterMatcher } from '../../src/matcher'
 import { MatcherLocation } from '../../src/types'
 import { mockWarn } from 'jest-mock-warn'
 import { components } from '../utils'
-import { ErrorTypes } from '../../src/errors'
 
 const currentLocation = { path: '/' } as MatcherLocation
 // @ts-expect-error
--- a/__tests__/matcher/addingRemoving.spec.ts
+++ b/__tests__/matcher/addingRemoving.spec.ts
@@ -402,31 +402,5 @@ describe('Matcher: adding and removing records', () => {
       )
       expect('same param named').not.toHaveBeenWarned()
     })
-
-    describe(`not appropriate values placed in components property"`, () => {
-      it('not appropriate object placed in components propery -> warn + error', () => {
-        try {
-          matcher.addRoute({
-            path: '/',
-            components: components.Home,
-            name: 'home',
-          })
-        } catch ({ message }) {
-          expect(message).toBe('Invalid route record "components" property.')
-          expect('should be object.').toHaveBeenWarned()
-        }
-      })
-
-      it('not appropriate lazyload in components property -> warn', () => {
-        matcher.addRoute({
-          path: '/',
-          components: { default: import('../utils') },
-          name: 'home',
-        })
-        expect(
-          'Promise instead of a function that returns a Promise.'
-        ).toHaveBeenWarned()
-      })
-    })
   })
 })
--- a/__tests__/matcher/addingRemoving.spec.ts
+++ b/__tests__/matcher/addingRemoving.spec.ts
@@ -1,4 +1,4 @@
-import { createRouterMatcher, RouterMatcher } from '../../src/matcher'
+import { createRouterMatcher } from '../../src/matcher'
 import { MatcherLocation } from '../../src/types'
 import { mockWarn } from 'jest-mock-warn'
 
@@ -7,13 +7,8 @@ const currentLocation = { path: '/' } as MatcherLocation
 const component: RouteComponent = null
 
 describe('Matcher: adding and removing records', () => {
-  let matcher: RouterMatcher
-
-  beforeEach(() => {
-    matcher = createRouterMatcher([], {})
-  })
-
   it('can add records', () => {
+    const matcher = createRouterMatcher([], {})
     matcher.addRoute({ path: '/', component, name: 'home' })
     expect(matcher.resolve({ path: '/' }, currentLocation)).toMatchObject({
       name: 'home',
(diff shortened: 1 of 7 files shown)


src/matcher/index.ts
```diff
@@ -71,11 +74,66 @@ export function createRouterMatcher(
...
+    }
+  }
+
   function addRoute(
     record: RouteRecordRaw,
     parent?: RouteRecordMatcher,
     originalRecord?: RouteRecordMatcher
   ) {
+    // for route records.components guard
+    routeComponentsOptionGuard(record)
```
posva, line 136: this should be guarded with `if (__DEV__)` (not inside of the function)

src/matcher/index.ts
```diff
@@ -71,11 +74,66 @@ export function createRouterMatcher(
     return matcherMap.get(name)
   }
 
+  function routeComponentsOptionGuard(record: RouteRecordRaw): never | void {
```
posva, line 77: the function should just receive the `components` option and any other relevant option like name or path and not be ran if other scenarios

__tests__/matcher/resolve.spec.ts
```diff
@@ -6,11 +6,13 @@ import {
   MatcherLocationRaw,
   MatcherLocation,
 } from '../../src/types'
-import { MatcherLocationNormalizedLoose } from '../utils'
+import {
+  MatcherLocationNormalizedLoose,
+  components as mockComponents,
```
posva, line 11:
```suggestion
  components as componentsStub,
```

__tests__/matcher/addingRemoving.spec.ts
```diff
@@ -411,5 +412,36 @@ describe('Matcher: adding and removing records', () => {
       )
       expect('same param named').not.toHaveBeenWarned()
     })
+
+    describe(`not appropriate values placed in components property"`, () => {
+      let matcher: RouterMatcher
+      beforeEach(() => {
+        matcher = createRouterMatcher([], {})
```
posva, line 419: let's just create a matcher in every test like above

__tests__/matcher/addingRemoving.spec.ts
```diff
@@ -411,5 +412,36 @@ describe('Matcher: adding and removing records', () => {
       )
       expect('same param named').not.toHaveBeenWarned()
     })
+
+    describe(`not appropriate values placed in components property"`, () => {
```
posva, line 416: let's remove this group and move all tests to the warn group

src/matcher/index.ts
```diff
@@ -71,11 +74,66 @@ export function createRouterMatcher(
...
+          RouteComponent | null | undefined | void
+        > = (rawComponent as Lazy<RouteComponent>)()
+        if (!('catch' in componentPromise)) {
+          warn(
+            `Components property in record  with path "${path}" is not valid. ` +
+              `Property "components" should be object. ` +
+              `ex> { components: { default: Home } }`
+          )
+          // should throw error becuase components are not object
+          throw new Error(`Invalid route record "components" property.`)
```
posva, line 124: This can be simplified a lot, we only need to check if `components` is a component and tell the user they seem to have a typo and it should be `component: MyView` or `components: { default: MyView }`

posva (changes requested):

src/matcher/index.ts
eddie0329: Thanks for review!

Shouldn't we use `path`? 

```javascript
    warn(
            `Components property in record  with path 👉 "${path}" is not valid. ` +
              `Property "components" should be object. ` +
              `ex> { components: { default: Home } }`
          )
```

I think showing path we be more helpful debug codes.

src/matcher/index.ts
eddie0329: Thanks for guidance! 😄

I am confused state `simplified`. I have tried to guard cases below:
 
```javascript
{
   components: component, // 1 regular component
   components: () => import('./MyPage'), // 2 lazy loaded component
   components: { render: () => h('div', {}, 'hello') } // 3 functional component
}
```

However, I have to guard only `case 1` (case below)?

```javascript
{
   components: component, // 1
}
```

eddie0329:
@posva  Thank you for reviewing codes. 

I refactored codes what you've requested and add comments what i confused.

src/matcher/index.ts
posva: yes

__tests__/matcher/addingRemoving.spec.ts
```diff
@@ -7,23 +7,26 @@ const currentLocation = { path: '/' } as MatcherLocation
 const component: RouteComponent = null
 
 describe('Matcher: adding and removing records', () => {
+  let matcher: RouterMatcher
+
```
posva, line 11: I meant the opposite of what you did 😅

eddie0329: 😅

eddie0329:
@posva  I just wondered why this PR isn't approved or merged. 

Is there somthing blocked or reproduce bugs, let me know.

__tests__/warnings.spec.ts
```diff
@@ -271,4 +271,20 @@ describe('warnings', () => {
       'It should be called exactly one time in each navigation guard'
     ).toHaveBeenWarned()
   })
+
+  describe(`not appropriate values placed in components property"`, () => {
```
posva, line 275: let's remove the describe but keep the it, it's not necessary

__tests__/warnings.spec.ts
```diff
@@ -271,4 +271,20 @@ describe('warnings', () => {
...
+    it('not appropriate object placed in components propery -> warn + error', () => {
```
posva, line 276:
```suggestion
    it('warns if components option is a single component', () => {
```

src/matcher/index.ts
```diff
@@ -71,11 +72,37 @@ export function createRouterMatcher(
...
+      ) {
+        warn(
+          `Components property in record is not valid. ` +
+            `Property "components" should be component: MyView or ` +
+            `components: { default: MyView }`
+        )
+        // same with navigation guards
+        throw new Error(`Invalid route record "components" property.`)
+      }
+    }
```
posva, line 96: let's just use the isRouteComponent (or similar name) to check if recordComponents is a component, and if that's the case, return `{ default: components }` to auto fix the issue during dev

src/matcher/index.ts
```diff
@@ -71,11 +72,37 @@ export function createRouterMatcher(
...
+  }
+
   function addRoute(
     record: RouteRecordRaw,
     parent?: RouteRecordMatcher,
     originalRecord?: RouteRecordMatcher
   ) {
+    // for route records.components guard
+    if (__DEV__) routeComponentsOptionGuard(record.components)
```
posva, line 105: then we could here change the `record.components` to the returned value
```suggestion
    if (__DEV__) {
      record.components = routeComponentsOptionGuard(record.components)
    }
```

posva (changes requested):
Let's keep this simpler and autofix only when components is a single component

src/matcher/index.ts
eddie0329: @posva Thanks for review. One Quick question about mentioned `isRouteComponent` 

Using isRouteComponent should be like below?

```javascript
  function routeComponentsOptionGuard() {
    for (const name in recordComponents) {
      const rawComponent = recordComponents[name]
      if (isRouteComponent(rawComponent) // ⭐️ ) {
                // ...
      }
    }
    return recordComponents;
  }

function isRouteComponent(
  component: RawRouteComponent
): component is RouteComponent {
  return (
    typeof component === 'object' ||
    'displayName' in component ||
    'props' in component ||
    '__vccOpts' in component
  )
}
```

How can I guard precisely the difference between `single component = {}` and `{ default: single component = {} }`?

eddie0329: I pushed what I worked. (please refer last commit wip: tag attached)

src/matcher/index.ts
```diff
@@ -462,4 +485,20 @@ function checkMissingParamsInAbsolutePath(
   }
 }
 
+/**
+ * Allows differentiating lazy components from functional components and vue-class-component
+ *
+ * @param component
+ */
+function isRouteComponent(
```
posva, line 493: Import it instead of redeclaring it

posva (changes requested):

src/matcher/index.ts
eddie0329: It is slightly different `isRouteComponent` declared in `navigationGuard`. 😃

```javascript
Object.keys(component).length === 0 // in matcher/index.ts
typeof component === 'object'  // in navigationGuards.ts 
```

eddie0329: It may cause confutsion i changed name of function. and slightly change logic of guard.

eddie0329: isRawRouteComponent is added. 

```javascript
function isRawRouteComponent(
  component: RawRouteComponent
): component is RouteComponent {
  return (
    typeof component === 'object' ||
    typeof component === 'function' || //  ⭐  This was added because of () => import('./Home.vue')
    'displayName' in component ||
    'props' in component ||
    '__vccOpts' in component
  )
}
```

eddie0329: It can't be the same as  `isRouteComponent declared in navigationGaurds`.

or i can change `isRouteComponent declared in navigationGuards` same as above

posva:
Thanks a lot for the help with this. After further testing I'm seeing too many false positives with this warning, so we are going to drop the feature.

PR #1278, opened 2022-01-23
docs(lazy-loading): manualChunks as proper link

simon04:
Allows users to click on it without extracting from the example source code and copying it to the clipboard.


posva:
Thanks but I prefer the current version

PR #1279, opened 2022-01-23
docs(sidebar): rename "Navigation Results"

simon04:
The original entry "Navigation Failures" is too narrow with respect to the page content. I was reading the first paragraphs and wondering how this is related to failures. Only a later section deals with failures.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -179,7 +179,7 @@ const config = defineConfig({
                 link: '/guide/advanced/extending-router-link.html',
               },
               {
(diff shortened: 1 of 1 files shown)


posva:
Thank you but I think Navigation Failure is more compelling to "navigation that result in users stay on the same page". Maybe you have ideas about refactoring a little bit the beginning at https://next.router.vuejs.org/guide/advanced/navigation-failures.html to make that clearer?

posva:
Closing due to inactivity. Please open a new issue with a reference to this one if you can follow up with more information.

PR #1286, opened 2022-01-25
Update redirect-and-alias.md

cp3hnu:
The Chinese document is wrong.


posva:
Thanks but the current version seems better

PR #1292, opened 2022-01-29
Update explanation on HTML mode

brunoalfred:
On the HTML mode section: i think it's good to specify  HTML mode and not history mode this will make the sentence sound more clear **When using HTML mode, the URL will look "normal," e.g. `https://example.com/user/id`. Beautiful!**

### DIFF
--- a/docs/guide/essentials/history-mode.md
+++ b/docs/guide/essentials/history-mode.md
@@ -39,7 +39,7 @@ const router = createRouter({
 })
 ```
 
-When using history mode, the URL will look "normal," e.g. `https://example.com/user/id`. Beautiful!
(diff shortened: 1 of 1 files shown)


docs/guide/essentials/history-mode.md
```diff
@@ -39,7 +39,7 @@ const router = createRouter({
 })
 ```
 
-When using history mode, the URL will look "normal," e.g. `https://example.com/user/id`. Beautiful!
+When using HTML mode, the URL will look "normal," e.g. `https://example.com/user/id`. Beautiful!
```
posva, line 42:
```suggestion
When using `createWebHistory()`, the URL will look "normal," e.g. `https://example.com/user/id`. Beautiful!
```

Issue #1277, opened 2022-01-23
Document `strict` option


posva:
The `strict` option allows to take into account trailing slashes and is disabled by default. [Here is an example](https://paths.esm.dev/?p=ABMMIPQg4AIMpgHtRkCkDscFEkQmQPADDqVuG4aBWkJmiXdVACQ.&t=/test#), try checking unchecked the option.

It should be documented in the https://next.router.vuejs.org/guide/essentials/route-matching-syntax.html

---

from https://github.com/vuejs/router/issues/1276

PR #1294, opened 2022-02-04
docs: document strict and sensitive option

cnschwarz:
- [x] Add section to documentation like requested in #1277  https://next.router.vuejs.org/guide/essentials/route-matching-syntax.html
- [x] Link documentation between new documentation and migration guide from Vue 2 router
- [x] Update examples in documentation to match the correct implementation and configuration in router
- [x] Update RouteRecordRaw API docs with `sensitive` and `strict`

Hi @posva 
I have two open points that I would need your feedback on:
- Is the example with setting` strict true` on the top level and setting `sensitive` on a specific route enough or should I add more examples? Or do you prefer a different example?
- In the corresponding closed issue it's mentioned that  documentation should also be on the API docs for the RouteRecordRaw - is this needed from your perspective? 

I will complete this change once I have your response. 

Thanks and regards!

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -849,17 +849,16 @@ Route record that can be provided by the user when adding routes via the [`route
 - **See Also**: [Passing props to Route Components](../guide/essentials/passing-props.md)
 
 ### sensitive
-- **Type**: `PathParserOptions` (Optional) 
+- **Type**: `boolean` (Optional) 
 - **Details**: 
 
-  Enables a more restrictive route matching. Makes the route case sensitive. `sensitive` can be passed on route level and now also when creating the router with `createRouter()` on the top level.
+  Makes the route matching case sensitive, defaults to `false`. Note this can also be set at a route level.
 
 ### strict
-- **Type**: `PathParserOptions` (Optional) 
+- **Type**: `boolean` (Optional) 
 - **Details**: 
 
-  Enables a more restrictive route matching. Disallows the check of an optional `/` at the end of the path. `strict` can be passed on route level and now also when creating the router with `createRouter()` on the top level.
-
+  Strictly checks the presence or absence of a trailing slash (`/`) at the end of the path. Defaults to `false` meaning that by default a route `/users` matches both `/users` and `/users/`. Note this can also be set at a route level.
 
 ### meta
 
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -848,6 +848,19 @@ Route record that can be provided by the user when adding routes via the [`route
 
 - **See Also**: [Passing props to Route Components](../guide/essentials/passing-props.md)
 
+### sensitive
+- **Type**: `PathParserOptions` (Optional) 
+- **Details**: 
+
+  Enables a more restrictive route matching. Makes the route case sensitive. `sensitive` can be passed on route level and now also when creating the router with `createRouter()` on the top level.
+
+### strict
+- **Type**: `PathParserOptions` (Optional) 
+- **Details**: 
+
+  Enables a more restrictive route matching. Disallows the check of an optional `/` at the end of the path. `strict` can be passed on route level and now also when creating the router with `createRouter()` on the top level.
+
+
 ### meta
 
 - **Type**: [`RouteMeta`](#routemeta) (Optional)
--- a/docs/guide/essentials/route-matching-syntax.md
+++ b/docs/guide/essentials/route-matching-syntax.md
@@ -80,15 +80,17 @@ const routes = [
 As mentioned in [the migration guide](../migration/index#removal-of-pathtoregexpoptions). `pathToRegexpOptions` and `caseSensitive` have been replaced with `sensitive` and `strict` options. Beside passing them per route, they can now also be directly passed when creating the router with `createRouter()`:
 ```js
 const router = createRouter({
-  history: createWebHistory("/"),
-  [
-    // will match /users and /users/posva
-    { path: '/users/:userId', sensitive: true },
-    // will match /users and /users/42
-    { path: '/users/:userId(\\d+)?' },
+  history: createWebHistory(),
+  routes: [
+    // will match /users/posva but not:
+    // - /users/posva/ because of strict: true
+    // - /Users/posva because of sensitive: true
+    { path: '/users/:id', sensitive: true },
+    // will match /users, /Users, and /users/42 but not /users/ or /users/42/
+    { path: '/users/:id?' },
   ]
   strict: true,
-});
+})
 ```
 
 ## Optional parameters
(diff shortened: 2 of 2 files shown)


docs/guide/essentials/route-matching-syntax.md
```diff
@@ -75,6 +75,22 @@ const routes = [
 ]
 ```
 
+## Sensitive and strict route options
```
posva, line 78:
```suggestion
## Sensitive and strict route options 

```

docs/guide/essentials/route-matching-syntax.md
```diff
@@ -75,6 +75,22 @@ const routes = [
...
+const router = createRouter({
+  history: createWebHistory("/"),
+  [
+    // will match /users and /users/posva
+    { path: '/users/:userId', sensitive: true },
+    // will match /users and /users/42
+    { path: '/users/:userId(\\d+)?' },
+  ]
+  strict: true,
+});
```
posva, line 90:
```suggestion
const router = createRouter({
  history: createWebHistory(),
  routes: [
    // will match /users/posva but not:
    // - /users/posva/ because of strict: true
    // - /Users/posva because of sensitive: true
    { path: '/users/:id', sensitive: true },
    // will match /users, /Users, and /users/42 but not /users/ or /users/42/
    { path: '/users/:id?' },
  ]
  strict: true,
})
```

posva (changes requested):
Thanks!

posva:
About your questions:

- Yes, 2 examples are better
- Yes, it should be in the API docs too if it isn't there already

docs/guide/essentials/route-matching-syntax.md
```diff
@@ -75,6 +75,23 @@ const routes = [
...
+```
+
```
posva, line 93:
```suggestion
```

cnschwarz:
@posva thanks for the quick reply and review then I will review the RouteRecordRaw API docs and extend them if it's not there yet + add another example where I set strict only on a specific route level

docs/api/index.md
```diff
@@ -848,6 +848,19 @@ Route record that can be provided by the user when adding routes via the [`route
 
 - **See Also**: [Passing props to Route Components](../guide/essentials/passing-props.md)
 
+### sensitive
+- **Type**: `PathParserOptions` (Optional)
```
posva, line 852: these are boolean

docs/api/index.md
```diff
@@ -848,6 +848,19 @@ Route record that can be provided by the user when adding routes via the [`route
...
 - **See Also**: [Passing props to Route Components](../guide/essentials/passing-props.md)
 
+### sensitive
+- **Type**: `PathParserOptions` (Optional) 
+- **Details**: 
+
+  Enables a more restrictive route matching. Makes the route case sensitive. `sensitive` can be passed on route level and now also when creating the router with `createRouter()` on the top level.
+
+### strict
+- **Type**: `PathParserOptions` (Optional)
```
posva, line 858: same

docs/api/index.md
posva, line 852:
```suggestion
- **Type**: `boolean` (Optional) 
```

docs/api/index.md
posva, line 858:
```suggestion
- **Type**: `boolean` (Optional) 
```

docs/api/index.md
```diff
@@ -848,6 +848,19 @@ Route record that can be provided by the user when adding routes via the [`route
 
 - **See Also**: [Passing props to Route Components](../guide/essentials/passing-props.md)
 
+### sensitive
+- **Type**: `PathParserOptions` (Optional) 
+- **Details**: 
+
+  Enables a more restrictive route matching. Makes the route case sensitive. `sensitive` can be passed on route level and now also when creating the router with `createRouter()` on the top level.
```
posva, line 855:
```suggestion
  Makes the route matching case sensitive, defaults to `false`. Note this can also be set at a route level.
```

docs/api/index.md
```diff
@@ -848,6 +848,19 @@ Route record that can be provided by the user when adding routes via the [`route
...
+
+### strict
+- **Type**: `PathParserOptions` (Optional) 
+- **Details**: 
+
+  Enables a more restrictive route matching. Disallows the check of an optional `/` at the end of the path. `strict` can be passed on route level and now also when creating the router with `createRouter()` on the top level.
```
posva, line 861:
```suggestion
  Strictly checks the presence or absence of a trailing slash (`/`) at the end of the path. Defaults to `false` meaning that by default a route `/users` matches both `/users` and `/users/`. Note this can also be set at a route level.
```

docs/api/index.md
```diff
@@ -848,6 +848,19 @@ Route record that can be provided by the user when adding routes via the [`route
...
+
+
```
posva, line 863:
```suggestion
```

docs/guide/essentials/route-matching-syntax.md
```diff
@@ -75,6 +75,36 @@ const routes = [
 ]
 ```
 
+## Sensitive and strict route options 
+
+As mentioned in [the migration guide](../migration/index#removal-of-pathtoregexpoptions). `pathToRegexpOptions` and `caseSensitive` have been replaced with `sensitive` and `strict` options. Beside passing them per route, they can now also be directly passed when creating the router with `createRouter()`:
```
posva, line 80:
```suggestion
By default, all routes are case-insensitive and match routes with or without a trailing slash. e.g. a route `/users` matches `/users`, `/users/`, and even `/Users/`. This behavior can be configured with the `strict` and `sensitive` options, they can be set both at a router and route level:

```

docs/guide/essentials/route-matching-syntax.md
```diff
@@ -75,6 +75,36 @@ const routes = [
...
+  history: createWebHistory(),
+  routes: [
+    // will match /users/posva but not:
+    // - /users/posva/ because of strict: true
+    // - /Users/posva because of sensitive: true
+    { path: '/users/:id', sensitive: true },
+    // will match /users, /Users, and /users/42 but not /users/ or /users/42/
+    { path: '/users/:id?' },
+  ]
+  strict: true,
```
posva, line 92:
```suggestion
  strict: true, // applies to all routes
```

docs/guide/essentials/route-matching-syntax.md
```diff
@@ -75,6 +75,36 @@ const routes = [
...
+const router = createRouter({
+  history: createWebHistory(),
+  routes: [
+    // will match /users/posva/ but not /Users/posva
+    { path: '/users/:id', sensitive: true },
+    // will match /users, /Users, and /users/42 but not /users/42/
+    { path: '/users/:id?', strict: true },
+  ]
+})
+```
```
posva, line 106:
```suggestion
```

posva (approved):
Thanks!

PR #1303, opened 2022-02-10
Update history-mode.md

mecograph:
Added example for AWS S3 via CloudFront


posva:
Thanks, but I'm going to refrain from adding any more examples on how to handle the SPA fallback on different servers because this got out of control in the previous docs.

mecograph:
That's a pitty, because a static website hosted on S3 is pretty common usecase for a SPA imho

PR #1311, opened 2022-02-15
docs: add redirect guard example

horuskol:
Added an example where the guard returns a route location object, to help clarify what can be returned from a navigation guard. This supplements the examples where the guard returns false or a simple string path.


docs/guide/advanced/navigation-guards.md
```diff
@@ -28,6 +28,15 @@ And can optionally return any of the following values:
...
 - A [Route Location](../../api/#routelocationraw): Redirect to a different location by passing a route location as if you were calling [`router.push()`](../../api/#push), which allows you to pass options like `replace: true` or `name: 'home'`. The current navigation is dropped and a new one is created with the same `from`.
 
+```js
+router.beforeEach(async (to, from) => {
+  if (!someConditional(to)) return {
+    name: 'AlternativeRoute',
+    replace: true
+  }
+})
+```
```
posva, line 38:
```suggestion
  ```js
  router.beforeEach(async (to, from) => {
    if (
      // make sure the user is authenticated
      !isAuthenticated &&
      // ❗️ Avoid an infinite redirect
      to.name !== 'Login'
    ) {
      // redirect the user to the login page
      return { name: 'Login' }
    }
  })
  ```
```

posva:
Thanks!

PR #1320, opened 2022-02-23
fix: a bug in playground,'this' could not be used in setup function and tries will not be modified

ShunjiChen:
Find a bug in playground,'this' could not be used in setup function and tries will not be modified.We can use tries as a ref object instead.


posva:
Thanks but this was intentional

PR #1321, opened 2022-02-23
Rephrase Regexp to regular expression

vanillajonathan:
`RegExp` is the name of the object in JavaScript. The word "regex" is more generic and a shorthand for "regular expression".


posva (changes requested):
Thanks! Let’s use regex and keep the word custom where it was used

vanillajonathan:
Alright, done.

I found the word custom to be superfluous.

posva (approved):

PR #1322, opened 2022-02-24
Update index.md

jewpaltz:
That entire part of the page does not belong on a *Getting Started* page.
The first clue that it is in the wrong place is that it needs to use many concepts that haven't been introduced yet.
More problematic is that you've combined two incompatible methodologies on the same page - CDN implementation & SFC - without any indication that the second concept requires a different setup.

Actually, you don't have any documentation on how to get setup when using SFCs. That is going to be my next submission.

These issues (and a few more) arise from the fact that the deleted section does not belong here. It belongs in section about programmatic navigation. Excerpts from it also belong in every section referencing imperative access to to the router object (params, etc.).

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -71,30 +71,3 @@ app.mount('#app')
 
 // Now the app has started!
 ```
-
-By calling `app.use(router)`, we get access to it as `this.$router` as well as the current route as `this.$route` inside of any component:
-
-```js
-// Home.vue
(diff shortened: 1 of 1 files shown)


posva:
Thanks, it's true this should be here but we shouldn't remove the whole thing. Ideally, we should move to a vite-based setup with SFCs so that would require rewriting the whole page in a very different way and we are already working on it.

Issue #1254, opened 2021-12-21
Empty nested route does not work with named parent


iamandrewluca:
### Version
4.0.12

### Reproduction link
[codesandbox.io](https://codesandbox.io/s/dawn-brook-ucsfs?file=/src/main.ts)

https://youtu.be/xFELsyUpNv8 (video is processing)

Sandbox is created from this repo: https://github.com/iamandrewluca/issue-vue-router/blob/main/src/main.ts

### Steps to reproduce
- Click `/projects/the-id` link, `Project` is not rendered
- Refresh coesandbox preview frame, `Project` is rendered

### What is expected?
When navigating to `/projects/the-id`,  `Project` component registered as `''` (empty path) is rendered as nested index path

### What is actually happening?
When navigating to `/projects/the-id`,  `Project` component registered as `''` (empty path) is not rendered as nested index path. But is rendered if the page is refreshed

---
This is seems to happend when parent route has `name`, if I move parent `name` to child (with empty path), then it works.

I simplified the example as much as possible, in my case I have some levels of nested routes.

In example below `{ path: '', component: ProjectOverview },` is not rendered

```ts
export const routes: RouteRecordRaw[] = [
  {
    path: '/',
    component: App,
    children: [
      { path: '/:pathMatch(.*)*', component: NotFound },
      { path: '', component: Dashboard },
      {
        path: 'projects',
        component: RouterView,
        children: [
          { name: 'project-list', path: '', component: ProjectList },
          {
            name: 'project-details',
            path: ':id',
            component: ProjectDetails,
            children: [
              { path: '', component: ProjectOverview },
              { name: 'project-tasks', path: 'tasks', component: ProjectTasks },
              {
                name: 'project-technologies',
                path: 'technologies',
                component: ProjectTechnologies,
              },
              { name: 'project-team', path: 'team', component: ProjectTeam },
              { name: 'project-edit', path: 'edit', component: ProjectEdit },
              {
                name: 'project-timeline',
                path: 'timeline',
                component: ProjectTimeline,
              },
            ],
          },
        ],
      },
    ],
  },
  {
    path: '/auth',
    component: Auth,
    children: authRoutes,
  },
]
```

posva:
This is expected behavior and you **do need** to put the name on the child:

```js
{
		path: '/projects/:id',
		component: RouterView,
		children: [{ name: 'project-details', path: '', component: Project }],
}
```

iamandrewluca:
[...]

PR #1339, opened 2022-03-12
docs: Added note about using named routes in nested routes

Jess-2021:
When I use named routes in nested routes, I have no idea about the why the empty nested path no work, same result after I read documentation about nested routes. Until i saw this issues #1254.
```js
const routes = [
  {
    name: 'UserRoute',
    path: '/user/:id',
    component: User,
    children: [
      // It can't show the `UserHome` component, when I link to User by named route.
      { path: '', component: UserHome },
    ],
  },
]
```
I thought if we could add some notes about using named routes in nested routes, it might help some guys avoid this problem.

### DIFF
--- a/docs/guide/essentials/nested-routes.md
+++ b/docs/guide/essentials/nested-routes.md
@@ -97,4 +97,19 @@ const routes = [
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


Jess-2021:
Good idea. I will have a try.

posva:
I created https://github.com/vuejs/router/issues/1376 to track this

PR #1360, opened 2022-04-05
docs: improve meta extending

matthiassommer:
The given documentation was not working.
To augment (and not replace) the existing RouteMeta interface of vue-router, one has to define a module.

### DIFF
--- a/docs/guide/advanced/meta.md
+++ b/docs/guide/advanced/meta.md
@@ -61,7 +61,9 @@ router.beforeEach((to, from) => {
 It is possible to type the meta field by extending the `RouteMeta` interface from `vue-router`:
 
 ```ts
-// Ensure this file is parsed as a module regardless of dependencies.
+// This can be directly added to any of your `.ts` files like `router.ts`
+// It can also be added to a `.d.ts` file, in which case you will need to add an export
+// to ensure it is treated as a module
 export {}
 
 import 'vue-router'
--- a/docs/guide/advanced/meta.md
+++ b/docs/guide/advanced/meta.md
@@ -58,10 +58,12 @@ router.beforeEach((to, from) => {
 
 ## TypeScript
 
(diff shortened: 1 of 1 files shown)


docs/guide/advanced/meta.md
```diff
@@ -58,10 +58,12 @@ router.beforeEach((to, from) => {
 
 ## TypeScript
 
-It is possible to type the meta field by extending the `RouteMeta` interface:
+It is possible to type the meta field by extending the `RouteMeta` interface from `vue-router`:
 
 ```ts
-// typings.d.ts or router.ts
+// Ensure this file is parsed as a module regardless of dependencies.
```
posva, line 64: Not sure if this is correct. I think you might need the `export {}` in a `.d.ts` file only

posva, line 64:
```suggestion
// This can be directly added to any of your `.ts` files like `router.ts`
```

docs/guide/advanced/meta.md
```diff
@@ -58,10 +58,12 @@ router.beforeEach((to, from) => {
...
+export {}
```
posva, line 65:
```suggestion
// It can also be added to a `.d.ts` file, in which case you will need to add an export
// to ensure it is treated as a module
export {}
```

docs/guide/advanced/meta.md
aluanhaddad: That is not correct. This is absolutely significant in `.ts` files. I wrote that comment on an answer on stack overflow about this very package. https://stackoverflow.com/a/63999522/1915893 

The point here is that, for a file to be considered a module, thus allowing it to contain a module augmentation, it must contain one or more ES Module `import` or `export` statements. Otherwise it will be treated as a script and not a module under the defaults of most tools.

PR #1375, opened 2022-04-19
docs: Added example for passing custom props

### LINKED ISSUE #1361: How to pass props to an component rendering via RouterView?
### Version
4.0.14

### Reproduction link
[codepen.io](https://codepen.io/maddycrowle/pen/yLpjdwQ)

### Steps to reproduce
Copy this little code and put it in any Vue3 component.

### What is expected?
Prop is passing to children component.

Extraneous non-props attributes (product) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.

---

I have code like:
```

<template>
 <router-view :an-prop="product" />
</template>
<script setup lang="ts">

...

const product: Ref<IProduct|undefined>: ref();

...

</script>
```

I need to pass product as a prop to the component which loads in router-view.

Some time ago it was possible just like `<router-view :an-prop="product" />`, but since new version of Vue was released I've got warning:

> Extraneous non-props attributes (product) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.

Maybe I'm blind, but I honestly didn't find anything about it in the official documentation, only trivial examples of using <router-view> and billions of trivial examples of using vue-router programmatically. After hours of experiments and studies search engines' results I found out that now router-view uses slots and end up with such code:

```
<router-view v-slot="{ Component }">
 <component :is="Component" />
</router-view>
```

Which gave me not much, actually. Anyway I don't understand what to do next. Is there any way to pass props to an component rendering via RouterView nowadays in vue3?

MilosPaunovic:
An example for passing custom props to `<router-view>` component.

Closes #1361

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -188,6 +188,17 @@ If you add a `target="_blank"` to your `a` element, you must omit the `@click="n
 - `Component`: VNodes to be passed to a `<component>`'s `is` prop.
 - `route`: resolved normalized [route location](#routelocationnormalized).
 
+### Example: Passing custom props to `<router-view>` component
+
+```html
+<router-view v-slot="{ Component }">
+  <component
+    :is="Component"
+    your-prop="your-value"
+   />
+</router-view>
+```
+
 ## createRouter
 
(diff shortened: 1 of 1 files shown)


posva:
Thanks, I ended up writing something quite different.

PR #1378, opened 2022-04-22
Update scroll-behavior.md

wangze0618:
[更改]第48行的“falsy” 改为 “false”


posva:
Thanks but this is correct!

Issue #266, opened 2020-05-22
Cannot read property of listen of undefined 


denistsoi:
### Version
4.0.0-alpha.12

### Reproduction link
[https://codesandbox.io/s/vue-router-v4-reproduction-l5fzm?file=/index.html](https://codesandbox.io/s/vue-router-v4-reproduction-l5fzm?file=/index.html)

### Steps to reproduce
How to Reproduce:
When the developer omits the history property when instantiating createRouter function, the above error is shown.

```
const router = createRouter({
  routes: [{
    path: "/",
    component: { template: "<div>Home</div>"}, 
  }, {
    path: "/path2",
    component: { template: "<div>path</div>"}
  }]
})
```

### What is expected?
Error should indicate that the history method is a required field (and suggest one of...)

However, the docs indicate the for client side applications, the default is only createWebHistory (unless this is SSR).

### What is actually happening?

> Uncaught TypeError: Cannot read property 'listen' of undefined
> at createRouter (vue-router.esm-bundler.js:2269)
> at main.js:16
> createRouter @ vue-router.esm-bundler.js:2269
> (anonymous) @ main.js:16
> 

---
How to resolve:
 
Include history property within createRouter options.

```
const routerHistory = createWebHistory()
const router = createRouter({
  history: routerHistory,
  routes: [{
    path: "/",
    component: { template: "<div>Home</div>"}, 
  }, {
    path: "/cv",
    component: { template: "<div>CV</div>"}
  }]
})
```

Suggestion:

The error log should include an improved message that the history field should be createWebHistory() or others that are supported. (Can not suggest as it is not included in tests)

May wish to include that in the docs.

posva:
It's already in the docs: https://github.com/vuejs/vue-router-next#breaking-changes-compared-to-vue-router3x

denistsoi:
Ok @posva 

I’ll look into adding a PR in the coming days, 

Just wondering (apart what other history methods are currently or will be supported in the next release?)

Updating docs should be a quick edit from my end if you’re cool with that.

denistsoi:
Oh wait you just added a doc update (sorry didn’t notice)

PR #1386, opened 2022-04-30
docs: modify how the API is used #266

iDestin:
Fix undefined errors caused by old API usage


posva:
Thanks but as mentioned in other issues and prs, this is intentional

iDestin:
> Thanks but as mentioned in other issues and prs, this is intentional

Oh sorry I don't know about this, but I think this will be some barrier for beginners, if you don't mind, can I know why? 😂

Issue #1306, opened 2022-02-13
"<router-view> can no longer be used directly inside <transition>" warning showing when router-view is not used directly inside transition.


codigovision:
### Version
4.0.12

### Reproduction link
[jsfiddle.net/8gxaptq2/](https://jsfiddle.net/8gxaptq2/)

### Steps to reproduce
Run and view warning in the console.

### What is expected?
No warning.

### What is actually happening?
The warning is displayed in the console.

---
I believe this is a bug because the warning specifically states that router-view can no longer be used **directly** inside, yet it is shown when not used directly inside, and also the router-view does work inside of a transition that is not its direct parent.

codigovision:
Thanks for confirming that it's not an intended warning in this case.

After digging into the warnDeprecatedUsage function where the warning is thrown, I believe the issue is that it only looks at the parent, which is the parent component, but not necessarily the parent dom element, so the parent subtree could contain other elements. 

What if the condition, after checking that the parent type is "KeepAlive" or contains "Transition", also checks that the parent subtree type is "RouterView"? That way it would ensure that the direct child of the "transition" or "keep-alive" is in fact the "router-view".

```js
function warnDeprecatedUsage() {
    const instance = getCurrentInstance();
    const parentName = instance.parent && instance.parent.type.name;
    const parentDirectChild = instance.parent && instance.parent.subtree.type.name;

  if (parentName && parentDirectChild &&
        (parentName === 'KeepAlive' || parentName.includes('Transition')) &&
            parentDirectChild === 'RouterView') {
        const comp = parentName === 'KeepAlive' ? 'keep-alive' : 'transition';
        warn(`<router-view> can no longer be used directly inside <transition> or <keep-alive>.\n` +
            `Use slot props instead:\n\n` +
            `<router-view v-slot="{ Component }">\n` +
            `  <${comp}>\n` +
            `    <component :is="Component" />\n` +
            `  </${comp}>\n` +
            `</router-view>`);
    }
}
```

posva:
I don't think that would be enough but feel free to give this a try. There are a few existing test cases in `warning.spec.ts`, if anybody wants to add this, give it a try by adding a new test and making sure old ones still pass. Note this might be a bit difficult to achieve without false positives. In that case, we will mark this as a won't fix as the warning still brings value given a lot of people is migrating to Vue Router 4

posva:
cases that combine KeepAlive, Suspense, etc. I tested locally and it didn't pass the tests.

[...]

PR #1436, opened 2022-06-09
Internal API usage fix for null https://github.com/vuejs/router/issues/1306

ASnow:
router just throw exception  when instance is null


posva:
this shouldn't be necessary and doesn't fix https://github.com/vuejs/router/issues/1306
Do you have a repro?

posva:
Closing for the moment.

PR #1437, opened 2022-06-10
chore: migrate `vitest` from `jest`

webfansplz:
vitest-vue-router-next


__tests__/guards/extractComponentsGuards.spec.ts
```diff
@@ -86,14 +86,14 @@ describe('extractComponentsGuards', () => {
...
 
-  it('throws if component is null', async () => {
-    // @ts-expect-error
-    await expect(checkGuards([InvalidRoute], 2)).rejects.toHaveProperty(
-      'message',
-      expect.stringMatching('Invalid route component')
-    )
-    expect('is not a valid component').toHaveBeenWarned()
-  })
+  // it('throws if component is null', async () => {
```
webfansplz, line 89: Comment here because it run failed 😅 ,need help~

posva:
How much faster is this compared to the jest setup?
Can you do this against the v4.1 branch? We moved to a monorepo  with pnpm

webfansplz:
> How much faster is this compared to the jest setup? Can you do this against the v4.1 branch? We moved to a monorepo with pnpm

It looks similar in speed.

Okay,i’ll do it in v4.1 branch.

webfansplz:
Move to #1438

PR #1438, opened 2022-06-12
chore: migrate `vitest` from `jest`

webfansplz:
vitest-router


posva:
This one got automatically close because I merged the v4.1 branch. Can you create a new PR against `main`. It would be nice if someone can get this a further read and see how faster it is

PR #1442, opened 2022-06-15
docs: replace redirect links if no need

byog:


posva:
Can you do the changes against the branch v4.1?

PR #1448, opened 2022-06-24
install latest vue-router version

tal7aouy:
Hi,
when i run ```yarn add vue-router@4``` or ```npm install vue-router@4``` and go to package.json it's looks like this.
```
  "dependencies": {
    "vue-router": "4"
  }
```
but when i run  ```yarn add vue-router``` or ```npm install vue-router``` it is looks like this.
```
  "dependencies": {
       "vue-router": "^4.0.16",
  }


posva:
Thanks but this is totally valid: https://jubianchi.github.io/semver-check/#/4/4.0.1
It makes sure the installed version is always the version 4 which is the one the docs were written for

Issue #1456, opened 2022-07-05
vue-router 4.1.x missing types export


lizthegrey:
### Version
4.1.1

### Reproduction link
[github.com](https://github.com/microsoft/TypeScript/issues/47792)

### Steps to reproduce
upgrade from vue-router 4.0.16 to vue-router 4.1.1 using Yarn 4 / TypeScript 4.7 in ESM mode

### What is expected?
dist/vue-router.d.ts is picked up as the types

### What is actually happening?
dist/vue-router.d.ts is not detected as the types, because it is not in the exports fields, only in the top level types field.

compile error:
[tsl] ERROR in /home/lizf/eve-roster/src/client/admin/SwaggerPane.vue.ts(4,26)
      TS7016: Could not find a declaration file for module 'vue-router'. '/home/lizf/eve-roster/.yarn/__virtual__/vue-router-virtual-6bddcd632c/0/cache/vue-router-npm-4.1.1-b947f4f193-a3fe6b4ed4.zip/node_modules/vue-router/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/vue-router` if it exists or add a new declaration (.d.ts) file containing `declare module 'vue-router';`

---
https://github.com/microsoft/TypeScript/issues/47792 is the tracking issue for this globally for TypeScript.

PR #1457, opened 2022-07-05
fix(types): adds ESM export of types

lizthegrey:
Fixes https://github.com/vuejs/router/issues/1456; see also https://github.com/microsoft/TypeScript/issues/47792

### DIFF
--- a/packages/router/package.json
+++ b/packages/router/package.json
@@ -8,6 +8,7 @@
   "types": "dist/vue-router.d.ts",
   "exports": {
     ".": {
+      "types": "dist/vue-router.d.ts",
       "node": {
         "import": {
           "production": "./dist/vue-router.prod.cjs",


posva (commented):
Thanks a lot!

lizthegrey:
Confirmed, https://github.com/eve-val/eve-roster/pull/1262 applied cleanly now.

karlhorky:
I think this may have introduced a problem with the types for the `node16` value for the `compilerOptions.moduleResolution` option in `tsconfig.json` - see [Are the types wrong](https://arethetypeswrong.github.io/):

> 
Imports of "vue-router" under the node16 module resolution setting when the importing module is ESM (its extension is .mts or .mjs, or it has a .ts or .js extension and is in scope of a package.json that contains "type": "module") resolved to CJS types, but ESM implementations.

Screenshot 2023-03-26 at 18 12 00

posva:
What is the problem? The tool isn't clear on whether this is a problem or not. If you have a boiled down reproduction showing a problem, that would help!

karlhorky:
Chatting with @andrewbranch (the author of Are The Types Wrong?), it appears that there is more context in the [readme on GitHub](https://github.com/arethetypeswrong/arethetypeswrong.github.io#:~:text=Types%20are%20CJS%2C%20but%20implementation%20is%20ESM).

I believe this is the relevant bit:

> - Types are CJS, but implementation is ESM. In the node16 resolution mode, TypeScript detects whether Node itself will assume a file is a CommonJS or ECMAScript module. If the types resolved are detected to be CJS, but the JavaScript resolved is detected to be ESM, this problem is raised. It is often caused by a dependency’s package.json doing something like:
> ```json
> {
>   "exports": {
>     "types": "./index.d.ts",
>     "import": "./index.mjs",
>     "require": "./index.js"
>   }
> }
> ```
> Because there is no "type": "module" setting here, the .js and .d.ts file will always be interpreted as a CommonJS module. But if the importing file is an ES module, the runtime will resolve to the .mjs file, which is unambiguously ESM. In this case, the module kind of the types misrepresents the runtime-resolved module. This tends to present the most issues for users when default exports are used. In the future, this tool may detect whether an export default might make this problem more severe and give a full explanation of why. In the meantime, you can read the explanation in https://github.com/microsoft/TypeScript/issues/50058#issuecomment-1404411380. The simple fix for the example above would be to add an index.d.mts file dedicated to typing the .mjs module, and remove the "types" condition.

posva:
I see, since we have no default export, it should be fine. I want to avoid having to add yet another file to the dist folder, it's a maintainer's nightmare 💀

karlhorky:
I think it's still indicating there is a problem here to be fixed. Often you can reproduce this with a simple StackBlitz with `Node16` module resolution.

karlhorky:
I'm not sure how to use `vue-router` in a standalone way (I'm just using some `import` I found in some example), but usually this is the type of errors that you would get:

StackBlitz demo: https://stackblitz.com/edit/node-ozajfl?file=tsconfig.json&file=index.ts%3AL6

```
This expression is not constructable.
  Type 'typeof import("stackblitz:/node_modules/vue-router/dist/vue-router")' has no construct signatures.
```

Screenshot 2023-04-28 at 14 48 56

karlhorky:
Ah, looking at [this file](https://github.com/vuejs/router/blob/7442d7f6b88d4da09ce42c22bc018377e8df0050/packages/playground/src/router.ts#L1-L21) in the playground, looks like I am doing things wrong.

Maybe this is more along the lines of a normal usage of `vue-router@^4` (no errors):

```ts
import { createRouter, createWebHistory } from 'vue-router';

const routerHistory = createWebHistory();
const router = createRouter({
  history: routerHistory,
  routes: [{ path: '/', component: null }],
});
```

Screenshot 2023-04-28 at 14 55 40


StackBlitz demo: https://stackblitz.com/edit/node-d5yw9g?file=tsconfig.json&file=index.ts

@andrewbranch is this a false positive of **Are The Types Wrong?** in this case?

Or is it:

- just coincidental that the CJS implementation matches the ESM implementation
- but there would be a good chance that the CJS implementation could drift, introducing bugs without any knowledge by maintainers

andrewbranch:
Here’s the concrete repro.

```ts
// Importing from ESM in Node
import VueRouter from 'vue-router';

const routerHistory = VueRouter.createWebHistory();
const router = VueRouter.createRouter({
  history: routerHistory,
  routes: [{ path: '/', component: null }],
});
```

Because the types indicate the module is CommonJS, a default import is allowed, and points to the `module.exports` object of the module. But the reality is that the module is ESM and has no default export, so this program will crash during Node’s module linking phase.

I always advocate for types to correctly represent their module kind, and consider this to be a problem worth fixing. But you can kind of see that you often have to go against conventional usage of the package in order to get yourself in real trouble with this issue.

posva:
Yeah, agree. I don’t think it’s worth complicating the build step and marking the package bigger to make an error for this. It’s already complicated enough 😆

PR #1468, opened 2022-07-11
feat(type): add navigation guard next callback type arguments

nandi95:
This is more of a discussion than a PR.

What's the best way of setting types for the page component?
Is there a module augmentation that would work here? (If so, maybe the docs can include a tip?)

This would be helpful when trying to set data on the component when using `beforeRouteEnter`: `next((vm) => { vm.myRef = fetchedData; });`

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,13 +27,13 @@
     "chalk": "^4.1.2",
     "enquirer": "^2.3.6",
     "execa": "^6.1.0",
-    "globby": "^13.1.1",
-    "lint-staged": "^13.0.0",
+    "globby": "^13.1.2",
+    "lint-staged": "^13.0.3",
     "minimist": "^1.2.6",
     "p-series": "^3.0.0",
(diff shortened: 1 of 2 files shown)


nandi95:
With this merge nothing has changed, the discussion is still relevant
https://github.com/vuejs/router/commit/0cf029859eb74561768606d49779ac12d2a99d39#diff-a8fafba408d30f77ae352e37581f57ce5477cf497beba3435f4fc137909f4b45R487
@posva

posva:
I never merged this but it seems like a77cd6b has the same commit hash as yours so GitHub thought this was merged.

To answer your question: there isn't a "best way to handle this". `beforeRouteEnter()`'s next callback is a last resort. Since there isn't much done inside the callback, typing the component as `any` is enough. You can indeed force the argument to `ComponentPublicInstance<P, B, D>` to have custom types 👍

PR #1494, opened 2022-08-02
feat(types): add option to extend `RouteRecordName`

nandi95:
This change allows for editing the `RouteRecordName` in the a module augmentation.

### TIP:
I have a route names map which I use to refer to names in router function.

```ts
export const routeNames = {
    public: {
        index: 'index',
        features: 'features',
        pricing: 'pricing',
        s: {
            index: 's-thing',
            edit: 's-thing-edit'
        }
    },
    app: {
        index: 'app',
    },
    fallback: 'all'
} as const;
```

Considering the above map.
In my app I created a little helper type:
```ts
type RouteNames<T extends Record<string, any>> = {
    [K in keyof T]: T[K] extends Record<string, any> ? RouteNames<T[K]> : T[K];
}[keyof T];
```
Which I could use in the module augmentation as if this PR gets merged
```ts
// eslint-disable-next-line @typescript-eslint/consistent-indexed-object-style
export interface RouteRecords extends Record<RouteNames<typeof routeNames>, never> {}
```

### DIFF
--- a/packages/router/src/types/index.ts
+++ b/packages/router/src/types/index.ts
@@ -205,9 +205,22 @@ export type RouteComponent = Component | DefineComponent
 export type RawRouteComponent = RouteComponent | Lazy<RouteComponent>
 
 /**
+ * Interface to add available RouteRecordName values
+ *
+ * @example
+ *    declare module 'vue-router' {
+ *        export interface RouteRecords {
+ *           [key: 'app' | 'contact-us' | 'services' | 'index']: never;
+ *        }
+ *    }
+ */
+export interface RouteRecords {
+}
+
+/**
  * Possible values for a user-defined route record's name
  */
-export type RouteRecordName = string | symbol
+export type RouteRecordName = keyof RouteRecords | string | symbol
 
 /**
  * @internal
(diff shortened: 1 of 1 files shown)


posva:
Thanks for the proposal! The direction we want to take is to take things a little bit further with params types as well 🙂 Take a look at https://github.com/posva/unplugin-vue-router and feel free to contribute there

nandi95:
> take things a little bit further with params types as well

That's great to hear!
Will this be part of the this router eventually?
Or the users will have to install that linked package? (Is that package is an experimental now and will be merged into this?)

posva:
Yes, the idea is to launch something official

andreasvirkus:
But wouldn't this solution benefit people
1. _right now_ without having to wait for the experimental plugin development
2. that do not wish to use the plugin

Could you reconsider reopening this @posva as it would allow for greater flexibility, e.g. currently you cannot have a simple `.startsWith()` or `.includes()` check because TS complains of those methods not existing on Symbol, even if your project never uses Symbols for route names and only sticks to strings.

PR #1495, opened 2022-08-02
docs: spelling fixes in doc blocks

nandi95:
Just some grammar fixes my IDE highlighted

### DIFF
--- a/packages/router/src/RouterLink.ts
+++ b/packages/router/src/RouterLink.ts
@@ -65,7 +65,9 @@ export interface RouterLinkProps extends RouterLinkOptions {
    */
   exactActiveClass?: string
   /**
-   * Value passed to the attribute `aria-current` when the link is exact active. Defaults to "page"
+   * Value passed to the attribute `aria-current` when the link is exact active.
+   *
+   * @default 'page'
    */
   ariaCurrentValue?:
     | 'page'
@@ -236,7 +238,7 @@ export const RouterLinkImpl = /*#__PURE__*/ defineComponent({
                 : null,
               href: link.href,
               // this would override user added attrs but Vue will still add
-              // the listener so we end up triggering both
+              // the listener, so we end up triggering both
               onClick: link.navigate,
               class: elClass.value,
             },
@@ -254,7 +256,7 @@ export const RouterLinkImpl = /*#__PURE__*/ defineComponent({
 export const RouterLink: _RouterLinkI = RouterLinkImpl as any
 
 /**
- * Typed version of the `RouterLink` component. Its generic defaults to the typed router so it can be inferred
+ * Typed version of the `RouterLink` component. Its generic defaults to the typed router, so it can be inferred
  * automatically for JSX.
  *
  * @internal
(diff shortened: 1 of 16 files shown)


packages/router/src/RouterLink.ts
```diff
@@ -65,7 +65,9 @@ export interface RouterLinkProps extends RouterLinkOptions {
    */
   exactActiveClass?: string
   /**
-   * Value passed to the attribute `aria-current` when the link is exact active. Defaults to "page"
+   * Value passed to the attribute `aria-current` when the link is exact active.
+   *
+   * @default 'page'
```
nandi95, line 70: I suspect there's more JSDoc improvements that can be made
https://jsdoc.app/tags-param.html#optional-parameters-and-default-values
but that can be done later

posva: We actually use tsdoc which can infer many things from types, so don't worry!
Can you change this to `defaultValue` instead? (https://tsdoc.org/pages/tags/defaultvalue/)

packages/router/src/history/common.ts
```diff
@@ -74,8 +74,8 @@ export type ValueContainer<T> = { value: T }
 export interface RouterHistory {
   /**
    * Base path that is prepended to every url. This allows hosting an SPA at a
-   * subfolder of a domain like `example.com/subfolder` by having a `base` of
-   * `/subfolder`
+   * sub folder of a domain like `example.com/sub folder` by having a `base` of
```
posva, line 77: let's keep the one word version, it makes more sense

packages/router/src/history/hash.ts
```diff
@@ -6,7 +6,7 @@ import { warn } from '../warning'
  * Creates a hash history. Useful for web applications with no host (e.g. `file://`) or when configuring a server to
  * handle any URL is not possible.
  *
- * @param base - optional base to provide. Defaults to `location.pathname + location.search` If there is a `<base>` tag
+ * @param {string} [base] - optional base to provide. Defaults to `location.pathname + location.search` If there is a `<base>` tag
```
posva, line 9: no need for these, let's keep the old version

packages/router/src/history/memory.ts
```diff
@@ -12,10 +12,10 @@ import {
 } from './common'
 
 /**
- * Creates a in-memory based history. The main purpose of this history is to handle SSR. It starts in a special location that is nowhere.
+ * Creates an in-memory based history. The main purpose of this history is to handle SSR. It starts in a special location that is nowhere.
  * It's up to the user to replace that location with the starter location by either calling `router.push` or `router.replace`.
  *
- * @param base - Base applied to all urls, defaults to '/'
+ * @param {string} [base=/] - Base applied to all urls
```
posva, line 18: same

packages/router/src/matcher/pathParserRanker.ts
```diff
@@ -52,23 +52,31 @@ export interface PathParser {
  */
 export interface _PathParserOptions {
   /**
-   * Makes the RegExp case sensitive. Defaults to false
+   * Makes the RegExp case-sensitive.
+   *
+   * @default false
```
posva, line 57: Same in this interface, let's use `defaultValue`

posva (changes requested):
Thanks a lot!

posva (approved):

Issue #1493, opened 2022-08-02
routing by "name" doesnt resolve nested routes with empty path


Schloemicher:
### Version
4.1.3

### Reproduction link
[codesandbox.io](https://codesandbox.io/s/intelligent-surf-psyjqo?file=/src/router.js)

### Steps to reproduce
- create a parent route with a name
- create children
  - one with empty string "" as path
  - one with an id

navigate somewhere by name to the parent 

### What is expected?
the component with the "" empty path is also shown

### What is actually happening?
only the parent is shown

---
it is somewhat unintuitive that you have to use the name of the child with the empty path because it doesnt work like the like de docs https://router.vuejs.org/guide/essentials/nested-routes.html promises

posva:
This is intended behavior so one can target the parent with no child. You even get a warning if you name the parent without naming the child route 🙂 . If you think this could be improved in documentation, please open a PR with the proposal

PR #1496, opened 2022-08-03
docs(nested-routes): add note for named routes

Schloemicher:
To avoid something like #1493

now it is clear why it works like this, clarified it in the docs as a reminder

### DIFF
--- a/packages/docs/guide/essentials/nested-routes.md
+++ b/packages/docs/guide/essentials/nested-routes.md
@@ -96,6 +96,36 @@ const routes = [
   },
 ]
 ```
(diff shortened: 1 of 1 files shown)


packages/docs/guide/essentials/nested-routes.md
```diff
@@ -96,5 +96,6 @@ const routes = [
   },
 ]
 ```
+**Note that if you use named routes you have to use the name of the child with the empty path to get the same effect**
```
posva, line 99: I think we can even add a short section _Named Nested Routes_

posva (changes requested):

posva (approved):
Thanks! I added a few extra changes

Issue #1270, opened 2022-01-16
Cant we expose removeHistoryListener api with createRouter


beiifeng:
### What problem does this feature solve?
Some time, we watn remove history listener manaul.

For example, in qiankun micro-front. Main-app and sub-app both vue, and beforeRoute guard with main-app. I passed the `history` and `router` of the main-app to the sub-app. The sub-app used the `history` of the main-app, and created the `router` itself, which caused the route guard in the main-app to fire twice

Example:
https://github.com/beiifeng/vue-router-next-report
Followe readme , from `首页` to `Home`,  then to `首页`，we find `BasicLayout:afterEach /Home` once in console. But find `BasicLayout:afterEach ` twice after we switch to `PppDemo1` .  `Sys2Demo1`  and `Sys2Demo2` is used api I want, only print once.

It's my commit:  https://github.com/beiifeng/vue-router-next/commit/b4cbbd0b80e3713aca45cb8cfa294b037aaa679c

### What does the proposed API look like?
```javascript
router = createRouter({
  history,
  routes,
});

router.removeHistoryListener();
```

beiifeng:
Cant we expose `setupListeners` too..

posva:
The sub app will automatically call `removeListeners()` when unmounted: https://github.com/vuejs/vue-router-next/blob/master/src/router.ts#L1214

The reproduction is **huge**, can you remove all unnecessary packages? At the very least everything related to linting, formatting, CSS ([you have both less and sass in there](https://github.com/beiifeng/vue-router-next-report/blob/main/packages/main/package.json#L36-L41)), TypeScript, and not having a mono repo would be better as well

We could expose `removeHistoryListeners()` or rather a `destroy()` method on the router like its history has but I want to make sure this is necessary

beiifeng:
I remove unuse file, and push it.

https://github.com/beiifeng/vue-router-next-report/blob/main/README.md

It maybe here https://github.com/single-spa/single-spa/blob/master/src/navigation/navigation-events.js#L105

beiifeng:
It looks like single-spa's error  https://github.com/single-spa/single-spa/issues/911

One way to resolve this error is expose history-listener api to let developers control listener manualy.
I want to know is this necessary too.

yiliang114:
It is a feature of single-spa, that it will fires PopStateEvent events when it wants to instruct all active applications to re-render.  [#1423](https://github.com/umijs/qiankun/issues/1423)

[...]

PR #1502, opened 2022-08-04
Critical Bug: Undefined URLs

brunomartinspro:
While using the latest version of vue and single-spa, i found that the url builder was adding undefined to the URL.
For me this happened when i went several levels below the current page.
This code avoids adding "undefined" to the URL when the "to" is undefined, when i want to return to the root of the SPA without doing a page refresh.

### DIFF
--- a/packages/router/src/history/html5.ts
+++ b/packages/router/src/history/html5.ts
@@ -287,7 +287,8 @@ function useHistoryStateNavigation(base: string) {
       )
     }
 
-    changeLocation(currentState.current, currentState, true)
+    if (currentState.current)
+      changeLocation(currentState.current, currentState, true)
 
     const state: StateEntry = assign(
       {},
--- a/packages/router/src/history/html5.ts
+++ b/packages/router/src/history/html5.ts
@@ -215,12 +215,19 @@ function useHistoryStateNavigation(base: string) {
      * base tag we can just use everything after the `#`.
      */
     const hashIndex = base.indexOf('#')
-    const url =
-      hashIndex > -1
-        ? (location.host && document.querySelector('base')
-            ? base
-            : base.slice(hashIndex)) + to
-        : createBaseLocation() + base + to
+
+    let url = ''
+
+    if (hashIndex > -1) {
+      if (location.host && document.querySelector('base')) {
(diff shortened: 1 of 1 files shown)


TCB13:
+1 on this, please merge.

brunomartinspro:
described in https://github.com/vuejs/router/issues/1507

posva:
Thanks for the PR but do you have a boiled-down repro or at the very least more information as to why or how `to` ends up being `undefined`?

brunomartinspro:
> Thanks for the PR but do you have a boiled-down repro or at the very least more information as to why or how `to` ends up being `undefined`?

It is really hard to replicate this in a new project (i am still trying), but i can provide more information in the meantime.
There are times when the currentState.current becomes undefined
image

posva:
That means either:

- Another lib or code is modifying the history and erasing the state: in which case the code should ensure the state persists so vue router can work
- Vue router is running while another router should be in charge: See https://github.com/vuejs/router/issues/1270

In both cases, the fix is not in vue router

brunomartinspro:
> That means either:
> 
> 
> 
> - Another lib or code is modifying the history and erasing the state: in which case the code should ensure the state persists so vue router can work
> 
> - Vue router is running while another router should be in charge: See https://github.com/vuejs/router/issues/1270
> 
> 
> 
> In both cases, the fix is not in vue router

@posva i completely understand your position, I will continue to try to identify the reason why the state has been somehow overridden by another MFE.

But when the history has been lost shouldn't we still navigate to the route? 

I added a commit to reflect this case.

posva:
> But when the history has been lost shouldn't we still navigate to the route?

The history cannot be _"lost"_. The router expects to be the only one interacting with it or to be manually deactivated when it's not active. In the issue mentioned above, people were also using single-spa and it seems things worked out for them, so you might want to ask them.

Good luck!

brunomartinspro:
@posva  After several days debugging i found out that this error happens when i go to an angular app and the return to a vue app.
Added the repository with the error, https://github.com/brunomartinspro/single-spa
I will continue to try to find a solution that does not require any change in the vue-router.

PR #1510, opened 2022-08-10
fix(types): normalized Query can have undefined properties

andresespinosapc:
When a query param is not present the query value is `undefined`, but the type is only `string | null`.

Here is the extract of code where I had this problem:
```
const route = useRoute();
const sortBy = ref(route.query.sort_by.toString())
```

I got `Cannot read property 'toString' of undefined` but the linter didn't give any errors.


posva (changes requested):
Thanks, I think this would indeed be better. Can you add a d-ts test and make sure others pass?

PR #1538, opened 2022-08-30
docs: code example on 'Fetching Before Navigation' with a better context

cr0wg4n:
Hi Vue team 😄  I was reading the documentation at https://router.vuejs.org/guide/advanced/data-fetching.html#fetching-before-navigation, and I found the code example quite confusing. I'm coming from version 2 of Vue, and `vm.setData()` method confused me. I've asked myself if that is a new syntax to modify the data component, but it's not. Maybe I'm not the only person that thinks like this. So, I added more context with the method setData in the example 😃. Thanks

### DIFF
--- a/packages/docs/guide/advanced/data-fetching.md
+++ b/packages/docs/guide/advanced/data-fetching.md
@@ -83,7 +83,7 @@ export default {
   },
   beforeRouteEnter(to, from, next) {
     getPost(to.params.id, (err, post) => {
-      // setData() is defined bellow (methods section)
+      // `setData` is a method defined below
       next(vm => vm.setData(err, post))
     })
   },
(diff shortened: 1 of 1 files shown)


packages/docs/guide/advanced/data-fetching.md
```diff
@@ -96,6 +96,15 @@ export default {
       this.error = error.toString()
     }
   },
+  methods: {
+    setData(error, post) {
+      if (error) {
+        this.error = error.toString()
```
posva, line 102:
```suggestion
        this.error = error
```

posva (changes requested):
Thanks, can you add a comment above the `vm.setData()` line saying it's a method defined below

cr0wg4n:
> 

Thanks :D, done

posva (approved):

PR #1543, opened 2022-08-31
fix: accept current navigation first before changing URL

FrankCheungCN:
**Issue:** 
Current route is not correct when location is changed.

**Demo:**
router v4：
https://codesandbox.io/s/vue-router-playground-forked-fsghsu?file=/src/main.js

router v3, which the current route is correct: 
https://codesandbox.io/s/vue-router-playground-forked-yi729z?file=/src/main.js


posva:
What is the actual use case behind monkey patching? Application code should use router.afterEach()

posva:
Yes: `router.afterEach()` is the place where reading the current location is ensured to be correct ✅

FrankCheungCN:
Thanks. I think there is no need to merge the change. I'll close it.

PR #1544, opened 2022-09-02
feat(types): improve link and view slots

xiaoxiangmoe:


packages/router/src/RouterLink.ts
```diff
@@ -269,7 +269,13 @@ export interface _RouterLinkI {
       RouterLinkProps
 
     $slots: {
-      default: (arg: UnwrapRef<ReturnType<typeof useLink>>) => VNode[]
+      default?: ({
+        route,
```
posva, line 273: I'm curious: does the destructuring help with something?
As in practice, this means we need to add a property if `useLink()` changes. If that's the case, can you add a `// NOTE: update `$slots` below` to the return statement of `useLink()`?

posva (commented):
Thank you!

packages/router/src/RouterLink.ts
xiaoxiangmoe: image

It will show all field of this interface

posva: so, better slightly better DX, got it 👍

posva (approved):

packages/router/src/RouterView.ts
```diff
@@ -20,6 +20,7 @@ import {
   RouteLocationNormalized,
   RouteLocationNormalizedLoaded,
   RouteLocationMatched,
+  RouteComponent,
```
posva, line 23:
```suggestion
```

PR #1547, opened 2022-09-07
Add previous route feature

nandi95:
Sometimes it would be nice to know what was the previous route. For example if I have a back button the route for that maybe either be the landing page if they their previous location wasn't this website otherwise the actual previous route or `.go(-1)` 

I'll finish the PR by adding documentation and testing if this feature is actually desired to be pulled in.

### DIFF
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -188,6 +188,12 @@ export interface Router {
    * @internal
    */
   // readonly history: RouterHistory
+
+  /**
+   * Previous {@link RouteLocationNormalized} or `null` if no previous exists.
+   */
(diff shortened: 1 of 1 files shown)


posva:
This was requested a long time ago at https://github.com/vuejs/vue-router/issues/883 but the discussion haven’t gone further. It would be useful to have real use cases for this, probably the RFC process would help too.

Today it’s already doable in user land:

```js
router.afterEach((to, from) => {
  to.from = from
})
```

(It can be added to the router or as a global ref too)

posva:
I think the discussion there is quite dead (it’s very old)😄

a short discussion in the rfcs repo would help move this forward

nandi95:
I commented on the linked issue. Not sure what you mean by rfcs repo.

PR #1557, opened 2022-09-21
docs: added notice for nginx root path

renegadevi:
If nginx cannot find index.html inherently. It will end up with access log spitting out ``rewrite or internal redirection cycle while internally redirecting to "/index.html"``.


posva:
Not sure this should be added as it looks more like nginx debugging

posva:
Closing due to inactivity

PR #1561, opened 2022-09-24
docs: Remove leading slash in documentation for apache redirect rule

bonham:
This PR is about a small glitch in the documentation about redirect rules when setting up with apache:

When deploying to a subfolder instead of root, the example from the documentation would not work since apache will redirect from `/subfolder/something` to `/index.html` instead of `/subfolder/something` to `/subfolder/index.html`

If the leading slash is removed as proposed, the example would work for both - deployment to root or to subfolder.


posva (approved):
Thanks!

PR #1563, opened 2022-09-28
chore: bronze sponsor link [skip ci]

andreasvirkus:
GitHub treats non-https-prefixed links as relative paths, resulting in 

https://github.com/vuejs/router/blob/main/www.vuejs.de

Screenshot 2022-09-28 at 09 48 17


posva:
Thanks!
@antony-k1208 I think you should update the link in your profile to include the https 😛 . I automatically pick the link from there

antony-k1208:
> 

Oh..my bad. Sorry I've just updated it 💪

PR #1581, opened 2022-10-17
docs: fix zh docs dynamic transition name mismatch error

Hellager:
In en docs they both being **meta.transitionName**
In zh docs, **route.meta.transition** mismatch with **to.meta.transitionName**


posva (changes requested):
Thanks, let's rename the other `route.meta.transition` as well then so all of them are `route.meta.transitionName`, even in English docs

PR #1594, opened 2022-11-03
docs: replace zh docs edit link text in chinese

suyanhanx:
Currently, the edit link text in zh docs is in english.
This PR fixes it.


packages/docs/.vitepress/config.ts
```diff
@@ -200,6 +200,7 @@ const config = defineConfig({
       '/zh/': {
         label: '中文',
         selectText: '选择语言',
+        editLinkText: '对本页提出修改建议',
```
posva, line 203:
```suggestion
        editLinkText: '为此页提供修改建议',
```

PR #1615, opened 2022-11-17
docs: typo

kentaro2014:
possible typo


posva (changes requested):
Thanks! This file is generated, can you change it in the router source code?

PR #1634, opened 2022-12-16
docs: improve sentence

gass-git:
"who" is used for people
"that" is used for things/objects

### DIFF
--- a/packages/docs/guide/advanced/meta.md
+++ b/packages/docs/guide/advanced/meta.md
@@ -5,7 +5,7 @@
   title="Learn how to use route meta fields"
 />
 
-Sometimes, you might want to attach arbitrary information to routes like transition names, who can access the route, etc. This can be achieved through the `meta` property which accepts an object of properties and can be accessed on the route location and navigation guards. You can define `meta` properties like this:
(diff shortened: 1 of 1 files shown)


packages/docs/guide/advanced/meta.md
```diff
@@ -5,7 +5,7 @@
   title="Learn how to use route meta fields"
 />
 
-Sometimes, you might want to attach arbitrary information to routes like transition names, who can access the route, etc. This can be achieved through the `meta` property which accepts an object of properties and can be accessed on the route location and navigation guards. You can define `meta` properties like this:
+Sometimes, you might want to attach arbitrary information to routes like transition names, that can access the route, etc. This can be achieved through the `meta` property which accepts an object of properties and can be accessed on the route location and navigation guards. You can define `meta` properties like this:
```
posva, line 8:
```suggestion
Sometimes, you might want to attach arbitrary information to routes like： transition names, or roles to control who can access the route, etc. This can be achieved through the `meta` property which accepts an object of properties and can be accessed on the route location and navigation guards. You can define `meta` properties like this:
```

PR #1656, opened 2023-01-08
docs: Update index.md (fix #1236)

### LINKED ISSUE #1236: addRoute in Promise/setTimeout do not get resolved on app mount
### Version
4.0.12

### Reproduction link
[jsfiddle.net/xch3or9a/](https://jsfiddle.net/xch3or9a/)

### Steps to reproduce
Set the registerInPromise to true/false to switch when the route is registered - instantly or in a promise. (might need to press Run first in the fiddle to match the url), also noticed that children aren't necessary to reproduce, breaks with root components too.

### What is expected?
Both cases should mount the Foo component.

Foo component is only mounted when registered synchronously, otherwise blank page is displayed.

---
If we name the routes, beforeEach guard also doesn't provide the named route as "to" argument when registered via Promise but gives a generic path route. Was sitting couple hours trying to figure out what's going on but couldn't tell exactly why this happens.

jayredk:
Fix #1236 

First time to work on Open Source project issue
If PR have any problem, please let me know
Thanks!

### DIFF
--- a/packages/docs/guide/index.md
+++ b/packages/docs/guide/index.md
@@ -74,7 +74,7 @@ app.mount('#app')
 // Now the app has started!
 ```
 
-By calling `app.use(router)`, we get access to it as `this.$router` as well as the current route as `this.$route` inside of any component:
+By calling `app.use(router)`, we are triggering the initial navigation and giving access to `this.$router` as well as the current route as `this.$route` inside of any component:
 
 ```js
 // Home.vue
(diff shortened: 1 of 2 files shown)


packages/docs/guide/index.md
```diff
@@ -73,6 +73,7 @@ app.mount('#app')
 
 // Now the app has started!
 ```
+**Note: Calling `app.use(router)` will triggers the initial navigation**
```
posva, line 76: Let's add it to the sentence below instead: `By caling `app.use(router)`, we are triggering the initial navigation and giving access to `this.$router` [...](rest of the sentence)

posva (changes requested):
Thanks!

PR #1668, opened 2023-01-21
Update installation.md : Add pnpm install

forinda:


skirtles-code:
I think this PR can be closed. That section of the docs has recently been reworked, causing conflicts with the change proposed here. The latest version of that page includes instructions for `pnpm`.

PR #1690, opened 2023-02-07
fix: modify the code comment

MonsterPi13:


posva:
Thanks but I think the current version makes more sense

PR #1691, opened 2023-02-08
 Maybe it's unnecessary to add this judgement

MonsterPi13:
The final element in the `toSegments` will always be empty or a string that is not `.` or `..`.

As a result, the `toPosition` will not increment when it exits the `for loop` and it will always be shorter than `toSegments' length`.


posva (approved):
All tests are passing, so I must have forgotten to remove them. Thanks a lot for noticing!

PR #1705, opened 2023-02-17
docs: fix bottom links' url

MonsterPi13:
Fix when i click bottom navigtor, it doesn't  redirect to the correct source code  of the current docs's page.


posva (approved):
Thanks!

PR #1727, opened 2023-03-08
chore: modify the way to get __dirname

140948940:
use dirname(new URL(import.meta.url).pathname) in node.js v18.15.0 have a bug,can't parse the directory correctly


posva (approved):
Thanks!

PR #1737, opened 2023-03-13
fix: Adjustment to NavigationGuardWithThis within dts file

daless14:
According to the documentation [here](https://router.vuejs.org/guide/advanced/navigation-guards.html#optional-third-argument-next), the "next" argument for navigation guards should be optional. Within the typescript dts file, this is typed as a required argument. Causes issues when attempting to type custom route guards with NavigationGuardWithThis.

### DIFF
--- a/packages/router/src/types/index.ts
+++ b/packages/router/src/types/index.ts
@@ -513,7 +513,7 @@ export interface NavigationGuardWithThis<T> {
     this: T,
     to: RouteLocationNormalized,
     from: RouteLocationNormalized,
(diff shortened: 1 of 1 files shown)


posva:
The current type is correct for the existing usage. What issue are you experiencing?

daless14:
Oh sorry about that. After your comment I looked at it again and realized I wasn't thinking about it correctly. I was trying to implement my beforeEach in a way that I could have multiple route guards but wasn't properly passing the arguments to my route guards. Thanks

PR #1741, opened 2023-03-15
chore: modify the way to get folders

140948940:
globby is have a bug in windows,can see this issues :https://github.com/sindresorhus/globby/issues/179. It's not a serious bug，But it makes learning to code tricky.

### DIFF
--- a/scripts/release.mjs
+++ b/scripts/release.mjs
@@ -344,6 +344,7 @@ async function getChangedPackages() {
     )
     lastTag = stdout
   }
+  // globby expects `/` even on windows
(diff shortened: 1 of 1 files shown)


scripts/release.mjs
```diff
@@ -344,7 +344,7 @@ async function getChangedPackages() {
     )
     lastTag = stdout
   }
-  const folders = await globby(join(__dirname, '../packages/*'), {
+  const folders = await globby((join(__dirname, '../packages/*').replace(/\\/g,'/')), {
```
posva, line 347:
```suggestion
  // globby expects `/` even on windows
  const folders = await globby((join(__dirname, '../packages/*').replace(/\\/g,'/')), {
```

PR #1755, opened 2023-03-26
chore: refactor checkFileSize function with async file reading 

kirklin:
Optimized the `checkFileSize` function in the `size-checks` script to improve its performance

### DIFF
--- a/scripts/check-size.js
+++ b/scripts/check-size.js
@@ -5,6 +5,10 @@ const { gzip } = require('zlib')
 const { compress } = require('brotli')
 
 async function checkFileSize(filePath) {
+  const stat = await fs.stat(filePath).catch(() => null)
+  if (!stat || !stat.isFile()) {
+    return
+  }
   const file = await fs.readFile(filePath)
   const minSize = (file.length / 1024).toFixed(2) + 'kb'
   const [gzipped, compressed] = await Promise.all([gzip(file), compress(file)])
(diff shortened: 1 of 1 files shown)


scripts/check-size.js
```diff
@@ -1,29 +1,38 @@
...
+    ])
+    const gzippedSize = (gzipped.length / 1024).toFixed(2) + 'kb'
+    const compressedSize = (compressed.length / 1024).toFixed(2) + 'kb'
+    console.log(
+      `${chalk.gray(
+        chalk.bold(path.basename(filePath))
+      )} min:${minSize} / gzip:${gzippedSize} / brotli:${compressedSize}`
+    )
+  } catch (err) {
+    console.error(err)
```
posva, line 23: let's remove the try/catch. It should crash the program if it throws

posva (changes requested):

scripts/check-size.js
```diff
@@ -1,18 +1,14 @@
-const fs = require('fs')
+const fs = require('fs').promises
 const path = require('path')
 const chalk = require('chalk')
-const { gzipSync } = require('zlib')
+const { gzip } = require('zlib')
 const { compress } = require('brotli')
 
-function checkFileSize(filePath) {
-  if (!fs.existsSync(filePath)) {
```
posva, line 8: can you add back the removed check?

posva (changes requested):

posva:
I think you have a checkbox to let maintainer edits unchecked

kirklin:
Oh, I accidentally closed it. Now I have reopened it.

PR #1756, opened 2023-03-26
docs: Update dynamic-matching.md

freemedom:
Is it necessary to add an explanation here? I just started learning vue and didn't understand this usage at first.


posva:
thanks! It's not necessary to add it as this is part of vue core docs

Issue #829, opened 2021-03-15
Documentation using modules


posva:
Right now the starting guide (https:/router.vuejs.org/guide/) still uses a vanilla index.html approach. I think that it was easier back when vue-router started but now it isn't anymore and that maybe that page should be updated using .vue components and pointing to vue CLI and vite. It should maybe also point to Vue docs explaining Vue files.

What are your thoughts on this?

Do you have any other ideas on how to restructure this section of the docs?

Stanley-SJTU:
https://www.vuemastery.com/blog/vue-router-a-tutorial-for-vue-3/
Found an example about how to use router in Vite.

ionous:
starting a new vue3 + vite project, one thing that confused me was how to import the router in order to use the code from the [getting started page](https://next.router.vuejs.org/guide/).  ( `import * as VueRouter from 'vue-router` seems to do the trick. )

the previous version's [installation page](https://router.vuejs.org/installation.html ) had an example import statement, and that was a nice help.

daniel-farina:
This was really painful, not having the imports needed for the examples on the router was quite confusing.

thedamon:
There have been several PRs to fix this rather obvious fault in the docs.. what do we need to do to get one in?

https://github.com/vuejs/router/pull/1951
https://github.com/vuejs/router/pull/1769
https://github.com/vuejs/router/pull/1323

posva:
@thedamon Most of them were actually incorrectly importing the router. It was better not to spread bad practices. Currently I need to focus on other stuff before getting back to the documentation, so feedback on difficult parts is welcome: https://github.com/vuejs/router/discussions/2068
BTW, https://github.com/vuejs/router/pull/1323 is the one that was the closest to what the page should look like but the author closed it probably because It's been 2 years

skirtles-code:
The examples have been updated by #2174.

I think this issue can be closed.

PR #1769, opened 2023-04-01
docs: improve getting started

Sombre-Osmoze:
- [x] Add router import in router documentation
- [x] Add tips to use `<router-view>…<router-view>` when use vite

### DIFF
--- a/packages/docs/guide/index.md
+++ b/packages/docs/guide/index.md
@@ -78,6 +78,10 @@ app.mount('#app')
 // Now the app has started!
 ```
 
+:::tip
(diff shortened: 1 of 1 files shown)


Sombre-Osmoze:
Hi, not sure if my documentation is explicite, when I tried to use vue-router with vite, I felt that some informations was missing to get the code to work (the mention of `router-view`)

Sombre-Osmoze:
Here is a screenshot of the changes: 
image

posva:
Thank you for the proposal but this is quite different from the direction we want to go (https://github.com/vuejs/router/issues/829)

PR #1772, opened 2023-04-03
chore: refactor size check

kirklin:


posva:
Thanks but this is fine how it is. It's only a size check script, it's not worth refactoring much as it doesn't have that much of an impact and is not changed often 😅

PR #1783, opened 2023-04-07
chore: remove unnecessary await keyword 

kirklin:
Summary
This PR removes the unnecessary await keyword from extractComponentsGuards and makes some minor adjustments.

### DIFF
--- a/packages/router/__tests__/guards/extractComponentsGuards.spec.ts
+++ b/packages/router/__tests__/guards/extractComponentsGuards.spec.ts
@@ -50,7 +50,7 @@ async function checkGuards(
   guardsLength: number = n
 ) {
   beforeRouteEnter.mockClear()
-  const guards = await extractComponentsGuards(
+  const guards = extractComponentsGuards(
     // type is fine as we excluded RouteRecordRedirect in components argument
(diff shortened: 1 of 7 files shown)


packages/router/src/router.ts
```diff
@@ -645,7 +645,7 @@ export function createRouter(options: RouterOptions): Router {
     const data: HistoryState | undefined = (to as RouteLocationOptions).state
     const force: boolean | undefined = (to as RouteLocationOptions).force
     // to could be a string where `replace` is a function
-    const replace = (to as RouteLocationOptions).replace === true
```
posva, line 648: this was intentional, see comment above

packages/router/src/devtools.ts
```diff
@@ -208,7 +208,7 @@ export function addDevtools(app: App, router: Router, matcher: RouterMatcher) {
       })
 
       router.afterEach((to, from, failure) => {
-        const data: TimelineEvent<any, any>['data'] = {
+        const data: TimelineEvent['data'] = {
```
posva, line 211: let's keep them

posva (changes requested):
Thanks!

posva (approved):

PR #1797, opened 2023-04-14
fix(warn): avoid redundant usage warn

leemove:
fix: https://github.com/vuejs/router/issues/1306

### DIFF
--- a/packages/router/__tests__/RouterView.spec.ts
+++ b/packages/router/__tests__/RouterView.spec.ts
@@ -411,6 +411,35 @@ describe('RouterView', () => {
       expect('can no longer be used directly inside').toHaveBeenWarned()
     })
 
+    it('does not warn if router-view is not used directly inside transition', async () => {
+      const route = createMockedRoute(routes.root)
+      const wrapper = mount(
+        {
+          template: `
+        <transition>
+          <div>
+            <router-view/>
+          </div>
+        </transition>
+        `,
+        },
+        {
+          props: {},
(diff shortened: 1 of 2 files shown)


posva (changes requested):
Can you add a test?

leemove:
> Can you add a test?

done

packages/router/__tests__/RouterView.spec.ts
```diff
@@ -411,6 +411,35 @@ describe('RouterView', () => {
       expect('can no longer be used directly inside').toHaveBeenWarned()
     })
 
+    it('does not warn if router-view is not used directly inside transition', async () => {
```
posva, line 414:
```suggestion
    it('does not warn if RouterView is not a direct-child of transition', async () => {
```

packages/router/src/RouterView.ts
```diff
@@ -240,9 +241,12 @@ export const RouterView = RouterViewImpl as unknown as {
 function warnDeprecatedUsage() {
   const instance = getCurrentInstance()!
   const parentName = instance.parent && instance.parent.type.name
+  const parentSubTreeType = instance.parent?.subTree?.type
```
posva, line 244:
```suggestion
  const parentSubTreeType = instance.parent && instance.parent.subTree && instance.parent.subTree.type
```

posva (approved):
Thanks! I will give this a local check later on and merge

PR #1810, opened 2023-04-21
docs: add tip for omitting `component` option since release of v4.1.0

KaygNas:


posva (changes requested):
Thanks. Instead there should be a section explaining this in the nested routes page (https://router.vuejs.org/guide/essentials/nested-routes.html). It shouldn't be a _tip_

KaygNas:
> Thanks. Instead there should be a section explaining this in the nested routes page (https://router.vuejs.org/guide/essentials/nested-routes.html). It shouldn't be a _tip_

Yes, maybe a new section about how to use nested routes to define a path structure would be better. A code example might be needed in the section, which demonstrate some routes with children might omit the `component` option and would just pass through.

But I am afraid that it is affordable to my English writing ability to complete such task...🤯

PR #1811, opened 2023-04-21
fix:enable vue-router to operate normally in the micro-front project with react-router instance

ygcedu:
When we create a front-end project with a microfront-end architecture like qiankun, the main application uses vue-router, and the sub app uses react-router, which causes value of history.state.current is undefined

### DIFF
--- a/packages/router/src/history/html5.ts
+++ b/packages/router/src/history/html5.ts
@@ -265,7 +265,7 @@ function useHistoryStateNavigation(base: string) {
     // Add to current entry the information of where we are going
     // as well as saving the current position
     const currentState = assign(
-      {},
+      {current: location.pathname},
       // use current history state to gracefully handle a wrong call to
       // history.replaceState
       // https://github.com/vuejs/router/issues/366


posva:
What is currently failing? Did you search existing issues about micro frontends? It's necessary for the router to work that the state is preserved correctly.
This looks like a duplicate of https://github.com/vuejs/router/pull/1219

ygcedu:
thank you for your reply.
The problem I encountered was that the main application used vue-router in qiankun, and the child application react-router call pushstate function, then I switch page through the main application's menu, the page cannot be updated because of error path，the path is http://localhost:3000undefined.
I see your reply of#1219, this issue just same with my problem, your ultimate solution is add a identity to control the router listener in&nbsp;beforeRoute guard?&nbsp;
This means that I have to modify the listening in the beforeEach function based on url?

&nbsp;


袁广超
***@***.***



&nbsp;




------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Eduardo San Martin ***@***.***&gt;; 
发送时间: 2023年4月21日(星期五) 下午5:15
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [vuejs/router] fix:enable vue-router to operate normally in the micro-front project with react-router instance (PR #1811)





 
What is currently failing? Did you search existing issues about micro frontends? It's necessary for the router to work that the state is preserved correctly.
 This looks like a duplicate of #1219

posva:
Yes, or usually something handled by the micro front end solution

PR #1824, opened 2023-05-01
Add translation status on page top

Jinjiang:
As a trial of a better translation & reading experience, in Vue Router docs as a sample, I added a label on the top of each page in a separated translation to show the hash and date of the checkpoint, which helps the readers know the translation status better.

@posva @NataliaTepluhina looks good?


wxsms:
Great idea. But will it lead to misunderstand in some case?

For example, some people raise a PR to update only one page. Since this badge looks like the same for all pages, will it still keep out of date before we make a valid sync?

posva (commented):
Thanks a lot! I refactored a few things. Do you think you include the usage of the script in the contributing docs?

Jinjiang:
> Great idea. But will it lead to misunderstand in some case?
> 
> For example, some people raise a PR to update only one page. Since this badge looks like the same for all pages, will it still keep out of date before we make a valid sync?

I think in this case the checkpoint still makes sense because we still need a comprehensive sync from there. For misunderstanding, we can change the diction a little. WDYT?

Jinjiang:
> Thanks a lot! I refactored a few things. Do you think you include the usage of the script in the contributing docs?

I haven't since I thought it was better to discuss it with you first. I'm struggling with the workflow since it requires a maintainer to manually update the translation status after step 5 (updating the checkpoint branch).

So this made me come up with this [translation workflow demo](https://github.com/Jinjiang/translation-workflow-demo). With the new workflow, no checkpoint branches or manual update step is required anymore.

If you feel the workflow in the demo is better, I can update this PR accordingly. 🙏

Jinjiang:
@posva I resolved the git conflict here and at the same time created another PR #1855 to practice [this new workflow](https://github.com/Jinjiang/translation-workflow-demo) as an alternative. I think we can decide to pick one of them to proceed. Feel free to take a further look. 🙏

Jinjiang:
I closed this PR and combined this into #1855. Let's continue the discussion there if you don't mind. 🙏

Issue #1846, opened 2023-05-12
Docs for API interfaces target incorrect type-aliases links


jordojordo:
### Reproduction

Follow type aliases links within documentation for interfaces, see steps below.

### Steps to reproduce the bug

1. Navigate to doc page for an interface - For example on the [`NavigationGuardNext`](https://router.vuejs.org/api/interfaces/NavigationGuardNext.html) page
2. Follow a link for a Type Alias - In the example follow the link for [`RouteLocationRaw`](https://router.vuejs.org/api/interfaces/NavigationGuardNext.html#Callable-NavigationGuardNext-Parameters_1) in the params definition.
3. Be redirected to `api/#<type>` - example: `api/#routelocationraw`

### Expected behavior

A given Type's link should redirect to the Type Alias header link for the type. In the `RouteLocationRaw` example it should redirect to `api/#Type-Aliases-RouteLocationRaw`.

### Actual behavior

Instead being redirected to `api/#Type-Aliases-<type>`, you are redirected to the top of the API documentation with the url being `api/#<type>`. As this header does not exist, there is no link to be directed to.

### Additional information

This happens in multiple locations within the API documentation. Examples:
- [`RouteLocation`](https://router.vuejs.org/api/interfaces/RouteLocation.html#Properties-name) definition for `name` mentions `RecordName` which leads to `api/#routerecordname`
- [`RouteLocationMatched`](https://router.vuejs.org/api/interfaces/RouteLocationMatched.html#Properties-children) definition for `children` points to `RouteRecordRaw` which leads to `api/#routerecordraw`
- [`Router`](https://router.vuejs.org/api/interfaces/Router.html#Methods) has multiple links in the "Methods" section which suffer the same fate.

thesheppard:
@posva I'm keen to assist with this 💪🏾

posva:
This ended up being fixed upstream 🎉

PR #1851, opened 2023-05-20
docs: Add correct type-aliases links for API interfaces

thesheppard:
Fix for #1846


posva (changes requested):
Api files are generated so this must be changed at the generation level 😅

PR #1854, opened 2023-05-21
Fix in docs for target incorrect type-aliases links #1846

SourabhRanka:
Fix for Docs for API interfaces target incorrect type-aliases links #1846


posva:
Thanks but these files are generated so the fix is within the generation

PR #1880, opened 2023-06-12
chore: defining default values for Props using withDefaults

Alfred-Skyblue:
Define props default values using withDefaults.


posva (approved):
Thanks!

PR #1903, opened 2023-06-27
Correct language to reflect pre-fetching

jmcooper:
The original statement that `Vue Router will only fetch [a lazily-loaded route/component] when entering the page for the first time` seems misleading since it does pre-fetch the route/component. My understanding is that the component chunks are actually pre-fetched by default.

### DIFF
--- a/packages/docs/guide/advanced/lazy-loading.md
+++ b/packages/docs/guide/advanced/lazy-loading.md
@@ -21,7 +21,7 @@ const router = createRouter({
 })
 ```
 
-The `component` (and `components`) option accepts a function that returns a Promise of a component and Vue Router **will only fetch it when entering the page for the first time**, then use the cached version. Which means you can also have more complex functions as long as they return a Promise:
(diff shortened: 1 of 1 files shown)


posva:
Thanks, but this depends on the build tool. Vue router doesn’t prefetch anything

jmcooper:
@posva That makes sense that this is a build tool concern, however, I think the confusion is that it isn't clear how the vue cli is configuring webpack out-of-the-box to work with lazy-loaded routes. Perhaps this is a question for the vue cli, not vue-router, but understanding why out-of-the-box, lazy-loaded routes are prefetched and how to disable it would be helpful. Back in 2018 there was a conversation about this where it was mentioned you could disable the prefetch plugin, but it seems that plugin is no longer being used so how to configure/disable prefetching for a lazy-loaded route is not clear.

jmcooper:
Nevermind, this was my error. When done correctly, it does not prefetch.

I was incorrectly using:
const ShoppingCart = import('../cart/ShoppingCart.vue')

When I should have been using:
const ShoppingCart = () => import('../cart/ShoppingCart.vue')

PR #1904, opened 2023-06-30
fix: handle undefined path in router resolve

kedrzu:
Suppose we execute for example:
```
router.replace({
    path: undefined,
    params: { foo: 'bar' }
);
```

Error would be thrown during `route.resolve` as we only check if `location` has property `path` but not if the path is defined and non null.

### DIFF
--- a/packages/router/__tests__/router.spec.ts
+++ b/packages/router/__tests__/router.spec.ts
@@ -319,6 +319,17 @@ describe('Router', () => {
     await router.push({ name: 'optional', params: {} })
   })
 
+  it('handles undefined path', async () => {
+    const { router } = await newRouter()
+
+    const route1 = router.resolve({
+      path: undefined,
+      params: { p: 'a' },
+    })
+    expect(route1.path).toBe('/')
+    expect(route1.params).toEqual({ p: 'a' })
+  })
+
   it('removes null/undefined optional params when current location has it', async () => {
     const { router } = await newRouter()
 
--- a/packages/router/src/matcher/index.ts
+++ b/packages/router/src/matcher/index.ts
@@ -290,7 +290,7 @@ export function createRouterMatcher(
       )
       // throws if cannot be stringified
       path = matcher.stringify(params)
-    } else if ('path' in location) {
+    } else if ('path' in location && location.path != null) {
       // no need to resolve the path with the matcher as it was provided
       // this also allows the user to control the encoding
       path = location.path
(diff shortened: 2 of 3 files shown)


kedrzu:
Yes, but it should not matter whether path is provided as `undefined` or not provided at all.
Especially, since in TypeScript `path?: string` property is satisfied by both passing `undefined` and not passing property value.

I found this bug when fixing the code recently migrated from Vue 2 to Vue 3. Thus, behavior in question is not backwards compatible. 

Also there is no readable error thrown. If the intention is really to require passing non-null path property or not passing it at, all it should be clearly stated in some exception being thrown. However, I doubt that's a good option, since adding some good smart-defaults is easy, better compliant with TS spec and compatible with previous version of `vue-router`.

posva:
Yea, TS does allow it but doesn’t allow to check for it with obj.property 🙃
I might change the types for that
Still, what is the code that ends up passing path: undefined?

kedrzu:
Piece of code looks like this:
```
const redirectPath = redirectFromCart ? localePathMain() : undefined;

if (queryChanged || redirectPath) {
  await router.replace({
    path: redirectPath,
    query,
  });
}
```

Basically, it redirects if path or query has been specified.
With requirement to not pass `path` at all I would have to replace it with multiple conditional clauses for all 3 cases:
- query changed and path is not null
- query changed and path is null
- query did not change and path is not null

This would look quite cumbersome. What I did to fix this issue, is I wrote a util function to get rid of undefined object props and used it like this:
```
if (queryChanged || redirectPath) {
    await router.replace(
        removeUndefinedProps({
            path: redirectPath,
            query,
        })
    );
}
```

Still, looks like a hack..

kedrzu:
And because most of the times in JS code `undefined` is not considered to be a value at all and there is usually no distinction between value being set to undefined or not being set at all.

I believe, that making this distinction for the sake of it does not make a lot of sense.

Also, please take a look at this piece of code, which does exactly the same, what I did by introducing this change: https://github.com/vuejs/router/blob/b480bd1d6ffe6a525e1869112e90a49322ecac02/packages/router/src/matcher/index.ts#L251

posva:
Sorry, I couldn't push some changes to the PR, I don't know why as the checkbox seems to be checked so I created https://github.com/vuejs/router/pull/1909. Feel free to make a review or even to create a new PR from there!

This could break some types for some people so I want to check a bit more before

PR #1907, opened 2023-07-01
Performance improvement by running the function in concurrent using `Promise.all` ✈️

sanjaiyan-dev:
This PR aims to enhance the performance of the function by using `Promise.all` to run it concurrently and utilize CPU resources more effectively. According to the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all), `Promise.all` returns a single promise that resolves when all of the promises passed as an iterable has resolved. This allows the function to execute multiple tasks in parallel and reduce the overall execution time.

Please review the code and let me know if you have any feedback or suggestions. I appreciate your time and consideration. Thank you.

### DIFF
--- a/scripts/release.mjs
+++ b/scripts/release.mjs
@@ -240,14 +240,14 @@ async function main() {
 
   step('\nCreating tags...')
   let versionsToPush = []
-  for (const pkg of pkgWithVersions) {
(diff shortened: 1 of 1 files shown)


posva:
Thanks but this is not needed in this script

PR #1927, opened 2023-07-12
Fix the wrong word. Update scroll-behavior.md

likeyoukang:
Fix the wrong word.
eg: falsy->false


posva:
Thanks but it’s correct

PR #1941, opened 2023-07-21
docs(ko): add ko-KR translation

niceplugin:
I would appreciate it if you could approve this PR so that more Koreans can see it. (Though the translation for the API section is not yet completed.)

(I have activated both the VueSchool links and Carbon ads.)

Please let me know if there's anything lacking!
Thank you.


niceplugin:
I have been maintaining the router site, just like the Pinia site, for over a year. Please let me know if there are any reasons why the translation site cannot be linked to the official site, or if there is anything lacking. Many Korean developers seem to be mistaking version 3 router for the latest router. (It appears they are misinterpreting `v3` as being the same as `Vue3` at https://v3.router.vuejs.org/kr/.)

posva (approved):
Thanks!

PR #1956, opened 2023-08-11
Added example for passing custom props

bjornharrtell:
I ended using the method described at https://github.com/vuejs/router/pull/1375 which was never merged into docs.

I think it's useful for cases where you derive some state in a parent (that might be reactive) that nested route components are dependent on.

I hope it's a supported case and that it deserves a more prominent place in the docs.

Thanks to @MilosPaunovic for the original snippet.


posva:
Thanks I had to merge locally

PR #1979, opened 2023-09-04
Link to current Apache configuration

Glandos:
2.2 is not maintained anymore. We could have used /current/, but it's not a permalink.


posva (approved):
Thanks!

Issue #1987, opened 2023-09-13
中文路径刷新时匹配不到路径


Preyou:
### Reproduction

无

### Steps to reproduce the bug

带有中文路径的路由

### Expected behavior

刷新后进入当前路由

### Actual behavior

进入404路由

### Additional information

在[这里](https://github.com/vuejs/router/blob/ccbdd5268175d1b222ca77d7a36d821ca4bef47e/packages/router/src/router.ts#L1233C1-L1233C1)，非英文会被编码，因此要先解码再匹配，这一行应该改成这样 `push(decodeURI(routerHistory.location))`

posva:
Don't open issues without reproductions. Thanks.

PR #1988, opened 2023-09-13
Update router.ts 修复非英文路径不匹配问题

Preyou:
#1987


posva:
This would just push unencoded urls and break

PR #1995, opened 2023-09-22
docs: lazy loading example

Kiansa:
just a suggestion to import inline so management of routes becomes easier.


posva (changes requested):
Thanks, can you add the example along the other route so there is only one `router`?

posva (approved):

PR #1998, opened 2023-09-26
Add z/Z to alphanumeric characters

mccahan:
Our friends at the end of the alphabet seem to have gotten left out, I assume unintentionally


posva (approved):
I don’t think the type is actually used but thanks 🤪

PR #2031, opened 2023-11-03
fix: keep optional params coming from a parent record

chuyuandu:
resolve TODO: only keep optional params coming from a parent record

### DIFF
--- a/packages/router/src/matcher/index.ts
+++ b/packages/router/src/matcher/index.ts
@@ -277,8 +277,8 @@ export function createRouterMatcher(
         paramsFromLocation(
           currentLocation.params,
           // only keep params that exist in the resolved location
-          // TODO: only keep optional params coming from a parent record
-          matcher.keys.filter(k => !k.optional).map(k => k.name)
+          // only keep optional params coming from a parent record
+          matcher.keys.filter(k => !k.optional).map(k => k.name).concat(matcher.parent ? matcher.parent.keys.filter(k => k.optional).map(k => k.name) : [])
         ),
         // discard any existing params in the current location that do not exist here
         // #1497 this ensures better active/exact matching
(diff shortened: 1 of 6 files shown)


packages/router/src/matcher/index.ts
```diff
@@ -277,8 +277,15 @@ export function createRouterMatcher(
         paramsFromLocation(
           currentLocation.params,
           // only keep params that exist in the resolved location
-          // TODO: only keep optional params coming from a parent record
-          matcher.keys.filter(k => !k.optional).map(k => k.name)
+          // only keep optional params coming from a parent record
+          matcher.keys
+            .filter(k => !k.optional)
+            .map(k => k.name)
```
posva, line 283: This could be done once at the end

posva (commented):
Thanks!

- It should also go for ancestors, not just direct parents.
- It should have tests about how params with matching names but coming from different routes are discarded

But I'm actually unsure if this is the desired behavior. Currently, matching params are silently passed around and people might find that useful. I wonder if making this behavior stricter will be better in the long term

packages/router/src/matcher/index.ts
chuyuandu: 1.  A direct parent's keys contains all optional params, includes ancestors
2. I'll open a new PR to improve the code if the behavior it is necessary, absolutely i think this is better to keep the optional params from parent record.

```js
{
    path: '/:type?/root',
    children: [
    {
        path: 'path_a',
        name: 'a',
    },
    {
        path: 'path_b',
        name: 'b',
    }
  ]
}

// current path:   '/optionType/root/path_a'

resolve({
    name: 'b'
});
```
I think `/optionType/root/path_b` is better than `/root/path_b`  in this case

posva: I guess o forgot how I wrote my own code 😂

chuyuandu: I suggest to keep optional params coming from a parent record,  that is really useful

chuyuandu:
@posva  hey, i have improved the code and add unit tests

posva:
thanks! I will check later on this year

posva (approved):

PR #2055, opened 2023-11-23
docs: update links to the core Vue docs

skirtles-code:
I've updated links to the docs for Vue core.

In most cases these were links to `v3.vuejs.org`. These had redirects configured and in most cases I just updated the link to reflect the location of the redirect.

For the Chinese translation I linked to `cn.vuejs.org` instead.

`history-mode.md` contained a link to `https://v3.vuejs.org/guide/ssr/introduction.html#what-is-server-side-rendering-ssr`. That link was broken and just led to the Vue homepage. I've updated it to point at the SSR guide, which is the closest match I could find. However, unlike the old SSR guide, the current SSR guide doesn't mention routing at all, so I'm not sure whether it's really equivalent.

### DIFF
--- a/packages/docs/guide/advanced/composition-api.md
+++ b/packages/docs/guide/advanced/composition-api.md
@@ -5,7 +5,7 @@
   title="Learn how to use Vue Router with the composition API"
 />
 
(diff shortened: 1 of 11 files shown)


posva (approved):
Thanks a lot!

PR #2056, opened 2023-11-25
docs: use vue-html for template examples

skirtles-code:
Using `vue-html` for the language of code blocks allows them to highlight Vue template syntax correctly. In particular, directive values are treated as JS expressions, rather than just HTML attributes:

Before
After

In a few cases I've also changed `vue` to `vue-html`. Using `vue` as the language requires the surrounding `<template>` tag, otherwise it won't be highlighted correctly. `vue-html` treats the whole block as template syntax, even without a `<template>` tag.

It isn't always clear whether an example should be treated as `vue-html` or `html`. Some examples are using in-DOM templates and only part of the HTML should be treated as a template. In those few examples I left it as `html`.

I've made the same changes to the Chinese translation.

### DIFF
--- a/packages/docs/guide/advanced/transitions.md
+++ b/packages/docs/guide/advanced/transitions.md
@@ -7,7 +7,7 @@
 
 In order to use transitions on your route components and animate navigations, you need to use the [`<RouterView>` slot](./router-view-slot):
 
-```html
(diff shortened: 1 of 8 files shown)


posva (approved):
Thanks!

PR #2069, opened 2023-12-05
docs: update custom transition example

skirtles-code:
The word `fallback` was removed by this change: https://github.com/vuejs/router/commit/950681aac7428cd91b47fd0b7a6aecb2b51c016f. This seems to have been unintentional.

I've reintroduced the word `fallback`. I also didn't think that `any` was quite the right word here, so I've made a couple of other tweaks to the wording.


posva (approved):
Thanks!

PR #2071, opened 2023-12-06
docs: expand installation.md

skirtles-code:
There were a few preliminaries that I felt were missing from the **Installation** page. I'm not sure how many people actually read that page, but at least the extra information will be explicitly stated for anyone who does.

I imagine this page will undergo significant changes when the docs are rewritten next year, but I thought it was worth making these additions now.

1. I've explicitly noted that `VueRouter` is exposed as a global object when using the global build.
2. I've added installation instructions for `pnpm`, alongside `npm` and `yarn`. To keep things concise, this is wrapped in a VitePress code group, similar to how commands like these are presented in the Vite docs.
3. I've tried to encourage newcomers to use `create-vue` instead.
4. I've added a note about how to access Vue Router via ES modules. I know this seems super obvious to anyone with JS experience, but the early docs examples are currently using the CDN approach and we get a steady stream of confused people on Vue Land who don't seem to understand the relationship between the global and ESM builds.

### DIFF
--- a/packages/docs/installation.md
+++ b/packages/docs/installation.md
@@ -12,14 +12,46 @@
 
 <!--/email_off-->
 
-## npm
+This will expose Vue Router via a global `VueRouter` object, e.g. `VueRouter.createRouter(...)`.
 
-```bash
+## Package managers
+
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks a lot!

PR #2080, opened 2023-12-14
Add pnpm

lazypolymath:


posva:
Thanks but we went with [this refactor](https://github.com/vuejs/router/pull/2071) instead

PR #2082, opened 2023-12-15
perf(RouterView): avoid parent rerenders when possible

### LINKED ISSUE #1701: View Component unnesseassarily update when child route changes
### Version
4.1.6

### Reproduction link
[codesandbox.io](https://codesandbox.io/s/clever-feather-cnpoqy?file=/src/views/Home.vue)

### Steps to reproduce
1. Open linked CodeSandbox
2. Observe Console
3. Click on "Home - A", and then on "Home - B"

### What is expected?
Should not log at all, because the component has receives no props and does not depend on any reactive data from the router.

So when a child route changes, it should not update.

Home.vue console.logs 3 times, meaning it re-rendered multiple times.

---
I'm pretty sure the reason for that is that `router-view` needs to re-render multiple times, and each time, a new `onVNodeUnmounted` callback is added to the VieComponent vnode's props object here:

https://github.com/vuejs/router/blob/e008551c168ed22c6007577541adf288382103ea/packages/router/src/RouterView.ts#L160-L170

So from the POV of Vue, the props changed. So it will run an update cycle of the router view.

posva:
Close #1701


Narretz:
Hi @posva  is this still relevant? Perf fixes are always good and I wonder  if this fixes a bug I'm seeing ...

posva:
@Narretz can you try using the pkg-pr published version?

Narretz:
@posva my problem was actually solved in a vue 3.5+ update

Narretz:
Hi @posva this has actually become relevant (to me) again, and as you can see from #1701, devs are still experiencing this as a problem and reporting it. I tested my case with the MR build and it fixes it.. So can this be merged and released? I guess there's a chance that some apps actually rely on the re-render behavior by now, but it still feels like a bug/unexpected behavior to me.

Narretz:
Hi @posva can you give this another look? See my comment above

posva:
This still requires more testing in different scenarios with Keep Alive, Transition, and Suspense to ensure everything keeps working. I haven't had the time to test all the combinations yet. There are other pending comments in the named issue. I will eventually get back to it.

PR #2085, opened 2023-12-17
docs: support dir in meta for pages

Jinjiang:
Support adding dir in meta like:

```md
---
dir: rtl
---
```

The demo screenshot

image

So we can control the direction per each page.

Thanks.

/cc @brc-dd

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -13,7 +13,7 @@
   },
   "dependencies": {
     "simple-git": "^3.18.0",
-    "vitepress": "1.0.0-rc.31",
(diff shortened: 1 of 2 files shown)


posva:
Shouldn't this be set directly in Vitepress theme 🤔

Jinjiang:
Yea, I think it's a good idea. Another PR has been created there. :-)

Jinjiang:
@posva I reset this branch and just bumped the vitepress version. It works as expected from my side. Please take a look. 🙏

posva (approved):

PR #2087, opened 2023-12-21
docs(fa): start Persian docs

the-pesar:


posva (changes requested):
Thanks a lot!

posva:
Thanks a lot! I’m still not sure if we want to merge all doc translations here. I will come back to this in January/february. In the meantime, we could just add an external link, like other docs. That could be merged quicker and give you more flexibility add changed until it’s ready

posva:
Closing due to inactivity

PR #2089, opened 2023-12-23
docs: add bun support to the Installation

the-pesar:


packages/docs/installation.md
```diff
@@ -32,6 +32,10 @@ yarn add vue-router@4
 pnpm add vue-router@4
 ```
 
+```bash [bun]
+bun install vue-router@4
```
skirtles-code, line 36: Shouldn't `bun add` be preferred to `bun install`?

posva:
Thanks but I'm starting to think it's not needed to add every package manager out there and I will likely reduce the list in the future. Bun users know how to use bun anyway

PR #2090, opened 2023-12-24
docs(cn): translate `Global injections within guards` section into Chinese

yionr:

PR #2091, opened 2023-12-24
docs(cn): translate `TypeScript` section of `Route Meta Fields` into Chinese

yionr:


packages/docs/.vitepress/translation-status.json
Jinjiang, line 1: Same as https://github.com/vuejs/router/pull/2090#discussion_r1436745733

Jinjiang (changes requested):

Jinjiang (approved):

Issue #2096, opened 2023-12-29
docs style error


heygsc:
### Reproduction

any page with TOC，such as : https://router.vuejs.org/guide/

### Steps to reproduce the bug

The right side cannot be fully displayed due to the banner.
<img width="1417" alt="image" src="https://github.com/vuejs/router/assets/103993866/4f79bea4-0fdd-4ac0-9559-b3a42f8a8527">

### Expected behavior

It can be fully displayed

### Actual behavior

The right side cannot be fully displayed due to the banner.

### Additional information

div with a class of "aside container", its padding-top is :
calc(var(--vp-nav-height) + var(--vp-layout-top-height, 0px) + var(--vp-doc-top-height, 0px) + 32px)

Perhaps a banner height should be added.

I have participated in submitting a commit before ( https://github.com/vuejs/docs/pull/2572 ) , and it was a similar bug. 
If possible, I hope to contribute again here.

Looking forward to the proposal of a solution.

posva:
Feel free to send a PR. It’s related to the banner in the top. I think there a css variable that needs to be set. The problem disappears when closing the banner

PR #2098, opened 2023-12-29
docs: style error caused by banner

heygsc:
fix: #2096


posva:
ooh I see now what you mean. Thanks a lot! I will try to merge once I'm done with the current refactor

posva (changes requested):
This doesn't seem to be the right fix 🤔 : the bottom right side is also offset and not visible with this (on Desktop)

heygsc:
> This doesn't seem to be the right fix 🤔 : the bottom right side is also offset and not visible with this (on Desktop)

Okay, I have created a new PR: #2099

PR #2099, opened 2024-01-03
docs: style error caused by banner

heygsc:
fix: #2096 

It seems that " --vt-banner-height " is not used in class " aside-container ".
And " --vp-layout-top-height " has no value in here, so I set it.

### DIFF
--- a/packages/docs/.vitepress/theme/components/VueMasteryBanner.vue
+++ b/packages/docs/.vitepress/theme/components/VueMasteryBanner.vue
@@ -269,7 +269,6 @@ onMounted(() => {
 
 <style>
 html.vuemastery-menu-fixed {
-  --vt-banner-height: 70px;
   --vp-layout-top-height: 70px;
 }
 html.vuemastery-menu-fixed .VPNav,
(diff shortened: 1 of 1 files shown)


heygsc:
It seems that " --vt-banner-height " is not used in class " aside-container ".
And " --vp-layout-top-height " has no value in here, so I set it.

packages/docs/.vitepress/theme/components/VueMasteryBanner.vue
```diff
@@ -270,6 +270,7 @@ onMounted(() => {
 <style>
 html.vuemastery-menu-fixed {
   --vt-banner-height: 70px;
+  --vp-layout-top-height: 70px;
```
posva, line 273:
```suggestion
  --vp-layout-top-height: 70px;
```

posva (approved):
Thanks a lot! Turns out the other variable does not exist anymore

posva:
@i5dr0id This is Worth noting for future banner updates on vitepress sites

PR #2053, opened 2023-11-22
fix(guards): run beforeRouteEnter with app context

posva:
Fix vuejs/router#2051

Issue #2051, opened 2023-11-22
`beforeRouteEnter` does not have an injection context if component is not yet loaded


rijenkii:
### Reproduction

https://jsfiddle.net/rv3p5ajn/13/

<details><summary>Code, if JSFiddle is down</summary>

```html
<div id="app">
  <pre>url: {{ $route.fullPath }}</pre>
  <ul>
    <li><router-link to="/page1">Page 1</router-link></li>
    <li><router-link to="/page2">Page 2</router-link></li>
  </ul>
  <router-view></router-view>
</div>

<script type="importmap">
  {
    "imports": {
      "vue": "https://unpkg.com/vue@3/dist/vue.esm-browser.js",
      "vue-router": "https://unpkg.com/vue-router@4/dist/vue-router.esm-browser.js",
      "@vue/devtools-api": "https://unpkg.com/@vue/devtools-api@6/lib/esm/index.js"
    }
  }
</script>

<script type="module">
import { createApp, hasInjectionContext } from "vue";
import { createRouter, createWebHistory } from "vue-router";

const Page1 = {
  template: `<div>page 1</div>`,
  beforeRouteEnter: () => console.log(hasInjectionContext()),
};
const Page2 = {
  template: `<div>page 2</div>`,
  beforeRouteEnter: () => console.log(hasInjectionContext()),
};

const router = createRouter({
  history: createWebHistory(),
  routes: [
    { path: "/page1", component: async () => Page1 },
    { path: "/page2", component: async () => Page2 },
  ],
});

const app = createApp({});
app.use(router);

window.vm = app.mount("#app");
</script>
```

</details>

### Steps to reproduce the bug

1. Open JSFiddle console
2. Click "Page 1"
3. Click "Page 2"
4. Click "Page 1" again
5. Click "Page 2" again

### Expected behavior

After all clicks value of `hasInjectionContext()` should be `true`.

### Actual behavior

On first and second clicks `hasInjectionContext()` is `false`. On third and fourth, when promises are already resolved, `hasInjectionContext()` is `true`.

### Additional information

According to the [docs](https://router.vuejs.org/guide/advanced/navigation-guards.html#Global-injections-within-guards), all guards should have an injection context.

All other guards do have an injection context, **but only before the first yield**.

I think the problem lies somewhere within `runWithContext`, because it behaves weirdly with promises:
```javascript
const { createApp, hasInjectionContext } = Vue;
const app = createApp({});

app.runWithContext(async () => {
[...]

PR #2117, opened 2024-01-26
fix(guards): run beforeRouteEnter with app context

yozi-developer:
Fix vuejs/router#2051
Close #2053

The PR is based on the https://github.com/vuejs/router/pull/2053


posva (approved):
Thanks!

Issue #2121, opened 2024-01-29
Vue router crashes when top level route target is wrapped in transition but has no child routes yet


robert-wloch-iits:
### Reproduction

For reproduction please see the provided Stackblitz project

### Steps to reproduce the bug

Here's a stackblitz project reproducing the issue: [Vue + Vue Router Error Reproduction](https://stackblitz.com/edit/vuejs-vue-router-with-transitions-error?file=README.md,src%2FApp.vue,src%2Frouter%2Findex.ts,src%2Fviews%2Froute-c%2FRouteCView.vue,src%2Fviews%2Froute-d%2FRouteDView.vue,src%2Fviews%2Froute-e%2FRouteEView.vue,src%2Fviews%2Froute-f%2FRouteFView.vue)

# Howto see the reproduction
First four steps show that transitions are working for Route A and Route B, even when going to Route C.
1. Navigate to "Works A"
2. Navigate to "Works B"
3. Navigate to "Works A"
4. Navigate to "Crashes C"
5. Navigate to "Works A" and see the the router views are not rendered anymore.

### Expected behavior

Vue router should silently ignore <ViewRouter></ViewRouter> when there are no children defined in the router config yet (iterative approach of software development). Even if it doesn't ignore them, navigation away to a previously working route should be possible without having to reload the application in the browser tab.

### Actual behavior

The top level router view in `App.vue` is wrapped in a `Transition` and for all top level router targets view components are there.
Three top level targets defined in their template a child `<ViewRouter></ViewRouter>` but the third, "Crashes C" has no children defined yet. When navigating to that top level target "Crashes C" and then to some other route, view router crashes.

Side-effect: When staying in Route "Crashes C" reloading the browser tab to restart the app will let a user stick in that route since navigating away from it will crash the vue router again.

### Additional information

Edit on 2024-02-01:
I've updated the StackBlitz to contain the Routes "Works D", "Crashes E" and "Works F" with the examples codes from my comments below.
Specifically "Crashes E" and "Works F" are a real mystery since the only difference is the missing commented out RouterView in the template section in "Works F". 

Edit on 2024-02-02:
With @K332's answer below I gave the [final section in the official documentation](https://router.vuejs.org/guide/advanced/transitions#Forcing-a-transition-between-reused-views) a try, which adds `:key="route.path"` to the inner `component` tag. Doing that, I now get to see the error in the browser console when navigating away from a "Crashes" route:
```
[...]

PR #2126, opened 2024-02-02
Update transitions.md

robert-wloch-iits:
The documentation update avoids situations described in #2121 and provides a link to a working demonstration project at Stackblitz.

### DIFF
--- a/packages/docs/guide/advanced/transitions.md
+++ b/packages/docs/guide/advanced/transitions.md
@@ -16,6 +16,7 @@ In order to use transitions on your route components and animate navigations, yo
 ```
 
 [All transition APIs](https://vuejs.org/guide/built-ins/transition.html) work the same here.
+If you want the transition to run on the first render, add `appear` as attribute to the transition tag.
 
 ## Per-Route Transition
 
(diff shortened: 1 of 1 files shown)


packages/docs/guide/advanced/transitions.md
```diff
@@ -72,13 +73,18 @@ router.afterEach((to, from) => {
...
 
+```js
+import { useRoute } from 'vue-router'
+const route = useRoute()
+```
+The `router-view` needs the route's full path as key. Although the route can be destructured from the slot props, that destructured route cannot be used in the key attribute. That's why getting the route from `useRoute` is the clean code solution. Setting a `key` on the inner `component` tag may cause application crashes in certain situations. See discussion in [issue 2121](https://github.com/vuejs/router/issues/2121) for more information about that.
+
 ```vue-html
-<router-view v-slot="{ Component, route }">
+<router-view v-slot="{ Component }" :key="route.fullPath">
```
posva, line 83: Thank you, but this is, most of the time, harmful in different ways:

- fullPath changes more than just path
- this doesn't work with transitions

On top of that, using `useRoute()` here breaks in other scenarios where we can force the resolved route for a router-view

robert-wloch-iits: What about just using `route.path`? Would it solve your concerns?

The point of my changes was to make it work with transitions. Could you please point out how it is not working? The linked Stackblitz project at the end of the PR is using transitions and they are all working.

posva: No, because the `key` needs to be on the component. Otherwise, the whole thing is being replaced. In your case, it works because of `appear`, but that shouldn't be necessary in the first place

posva: I will close this for now but feel free to open a new PR with a fix for #2121. Note there shouldn't be a need to change the current way of doing things.

PR #2127, opened 2024-02-04
docs: omitting parent components for nested routes

skirtles-code:
close #1810

The change to `index.ts` is intentional. Extending a theme in VitePress is now done using `extends` rather than `...`. The main difference is that `extends` will ensure `enhanceApp` is called on `DefaultTheme`. This is needed for the `<Badge>` to work.

The example is based on a similar example given at <https://github.com/vuejs/router/pull/1397>.

### DIFF
--- a/packages/docs/guide/essentials/nested-routes.md
+++ b/packages/docs/guide/essentials/nested-routes.md
@@ -148,4 +148,4 @@ const routes = [
 ]
 ```
 
-As the parent doesn't specify a route component, the top-level `router-view` will skip over the parent and just use the component from the relevant child instead.
+As the parent doesn't specify a route component, the top-level `<router-view>` will skip over the parent and just use the component from the relevant child instead.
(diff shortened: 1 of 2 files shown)


packages/docs/guide/essentials/nested-routes.md
```diff
@@ -128,3 +128,24 @@ const routes = [
...
+    children: [
+      { path: '', component: AdminOverview },
+      { path: 'users', component: AdminUserList },
+      { path: 'users/:id', component: AdminUserDetails },
+    ], 
+  },
+]
+```
+
+As the parent doesn't specify a route component, the top-level `router-view` will skip over the parent and just use the component from the relevant child instead.
```
posva, line 151:
```suggestion
As the parent doesn't specify a route component, the top-level `<router-view>` will skip over the parent and just use the component from the relevant child instead.
```

posva (approved):
Thank you!

PR #2130, opened 2024-02-06
docs: explain per-route guards on nested routes

skirtles-code:
While working on #2127, I had originally intended to include a note about per-route guards, but the specific functionality I was trying to reference doesn't appear to be documented, so I left it out.

This PR attempts to address that.

Specifically, when using nested routes, the `beforeEnter` per-route guard can be placed on either the parent or the child route. There is some subtlety to exactly when those guards will be called, which I believe is intentional and worth documenting.

I've also moved the final paragraph of that section into a `tip`. The new content about nested routes left that final paragraph feeling like a strange jump to a separate topic, and I think the `tip` box makes that less confusing.

### DIFF
--- a/packages/docs/guide/advanced/navigation-guards.md
+++ b/packages/docs/guide/advanced/navigation-guards.md
@@ -197,7 +197,28 @@ const routes = [
 ]
 ```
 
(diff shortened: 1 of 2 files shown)


posva (approved):
Great, thanks a lot!

Issue #2132, opened 2024-02-06
Router creation performance degradation when using many routes in comparison to v3


BobbieGoede:
### Reproduction

https://stackblitz.com/edit/github-afjnf8?file=src%2Frouter.ts

### Steps to reproduce the bug

* Run project and test the performance using `ab` or `siege` and compare the results.
  * I tested with the following command `siege -t 10s -c 100 "http://localhost:3000/"` to see the amount of requests within 10s using 100 concurrent users.
* Change the `dirCount` variable to adjust the amount of routes generated/used (default `100`, which generates 1101 routes). and test again.

### Expected behavior

Expected the amount of routes to have an impact on performance comparable to router v3. The performance difference is noticeable for SSR as a new router instance is created on each request.

### Actual behavior

Router creation doesn't scale as well as it did in v3 (presumably, I have no clean reproduction with v3 Router alone), please view the result comparison in this issue https://github.com/nuxt/nuxt/issues/25612 which compares Nuxt 2 and Nuxt 3.

If you have a clean/plain SSR project using router v3 I would be happy to include a reproduction with a performance comparison between v3 and v4, I tried making one but had no luck.

### Additional information

While debugging performance issues for `@nuxtjs/i18n` narrowed it down to the difference in performance between Nuxt 2 and Nuxt 3, specifically the performance impact of having a high amount of routes. Related issue here: https://github.com/nuxt/nuxt/issues/25612

Due to the way i18n routing works (same as in i18n for Nuxt 2), routes are generated/multiplied by the amount of configured languages. Of course I will look into alternative ways to make routing more efficient/performant when using i18n, but it's worth looking into why the performance of router creation has gotten worse and how we can improve it. 

My first guess would be that [`insertMatcher`](https://github.com/vuejs/router/blob/4d3cbf58295055ce045efda13db6a1afd144d172/packages/router/src/matcher/index.ts#L225-L240) doesn't scale very well, I couldn't find its equivalent in v3.

posva:
Thanks for the investigation.

It would be interesting to see an actual perf benchmark (without nuxt). The difference in creation shouldn't be that big. That said, Vue Router 3 had no path sorting like v4. It would also allow to have a flame graph but I think you are right with your guess of `insertMatcher` 😄 

Initially, I designed the router to accept a custom matcher implementation (e.g. a trie) that would be more performant but with fewer features than v3. I never got the time to actually work on it, though.

[...]

PR #2136, opened 2024-02-08
feat: router matcher sort cache

BobbieGoede:
* Related #2132 

This PR adds the `createRouterMatcherSortCache` function which accepts a sorted `RouterRecord` array (returned from `router.getRoutes()` and returns a map of keys and sort index. This can be passed as `restoreCache` option to `createRouter`, if this cache is provided the path sorting is skipped when adding routes during matcher creation and the sort cache is used to restore the order after all initial routes have been added.

The sort cache and restoration is not perfect, it's more of a proof of concept, I tried to add this functionality without making any breaking changes. 

The usage would be as follows:

```ts
import { createRouter, createMemoryHistory, createRouterMatcherSortCache } from "vue-router";

// during build or startup
const initialRouter = createRouter({
  history: createMemoryHistory(),
  routes,
});
const sortCache = createRouterMatcherSortCache(initialRouter.getRoutes());

// store and retrieve `sortCache` however you like ...

// on following router creations
const router = createRouter({
  history: createMemoryHistory(),
  routes,
  sortCache,
});
```

So far with limited testing the cache may have issues with 'duplicate' routes, there may not be enough information from records returned from `getRoutes` to create a fully unique key, though so far these duplicates have the same score and I don't know if the order between these have any significance. I think if the router matcher were exposed, or the matchers array directly (something like `getMatchers`), it may be possible to more accurately create a sort cache, I opted not to this in this iteration to keep changes to a minimum.

I have added a branch to my performance comparison which also tests router creation with cached sort [here].(https://github.com/BobbieGoede/vue-router-perf-bench/tree/feature/sort-cache-comparison). Just caching the order of the matchers greatly reduces the creation time, results on my machine:

| version/routes| 11       | 110      | 550      | 1100      | 2200      |
|---------------|----------|----------|----------|-----------|-----------|
| RouterV3      | 5.88ms   | 21.90ms  | 100.11ms | 200.38ms  | 405.75ms  |
| RouterV4      | 10.72ms  | 40.35ms  | 601.14ms | 2386.12ms | 9649.10ms |
| RouterV4Cached| 5.26ms   | 32.52ms  | 141.46ms | 275.60ms  | 553.46ms  |

This is just one of the ways router performance could be improved, I will still look into what it would like to support custom matchers and perhaps more accurate ways of precompiling the matcher. Maybe there are obvious flaws in this implementation, I'm not attached to this implementation so please let me know 😄!

### DIFF
--- a/packages/router/src/index.ts
+++ b/packages/router/src/index.ts
@@ -1,7 +1,7 @@
 export { createWebHistory } from './history/html5'
 export { createMemoryHistory } from './history/memory'
 export { createWebHashHistory } from './history/hash'
-export { createRouterMatcher } from './matcher'
+export { createRouterMatcher, createSortCache } from './matcher'
 export type { RouterMatcher } from './matcher'
 
 export { parseQuery, stringifyQuery } from './query'
--- a/packages/router/src/index.ts
+++ b/packages/router/src/index.ts
@@ -1,7 +1,7 @@
 export { createWebHistory } from './history/html5'
 export { createMemoryHistory } from './history/memory'
 export { createWebHashHistory } from './history/hash'
-export { createRouterMatcher, createSortCache } from './matcher'
+export { createRouterMatcher, createRouterMatcherSortCache } from './matcher'
 export type { RouterMatcher } from './matcher'
 
 export { parseQuery, stringifyQuery } from './query'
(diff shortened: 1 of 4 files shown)


BobbieGoede:
This is no longer relevant as better approaches are being discussed and merged! 💪🚀

PR #2138, opened 2024-02-11
test: earlier routes should have matching priority

skirtles-code:
My initial attempt at #2137 passed all the existing unit tests, but with hindsight it fails to take account of various important cases.

This PR attempts to add test cases for one of those problem cases. e.g. Consider the following code:

```js
routes: [
  {
    path: '/user/:id(1\\d*)',
    component: {}
  }, {
    path: '/user/:id(\\d+)',
    component: {}
  }
]
```

The path ranking gives these routes the same score. For a path of `/user/1234`, both would be eligible to match, but we would want the first route to match. The ordering matters, the earlier route needs to be given priority.

I'm not clear whether this is documented anywhere, but I think it's a reasonable assumption and something that people are likely to be relying upon, even though they may not realise it.

My attempt at using a binary search in #2137 to sort the routes broke this assumption, leading to routes with equal scores being in an essentially random order. I believe that PR is fixable, but these extra tests can be merged independently.

### DIFF
--- a/packages/router/__tests__/matcher/addingRemoving.spec.ts
+++ b/packages/router/__tests__/matcher/addingRemoving.spec.ts
@@ -418,6 +418,64 @@ describe('Matcher: adding and removing records', () => {
     )
   })
 
+  it('should give priority to earlier routes', () => {
+    const matcher = createRouterMatcher([], {})
+    matcher.addRoute({ path: '/:id(123\\d*)', component, name: 'first' })
+    matcher.addRoute({ path: '/:id(12\\d*)', component, name: 'second' })
+    matcher.addRoute({ path: '/:id(1\\d*)', component, name: 'third' })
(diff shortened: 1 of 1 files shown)


posva:
> I'm not clear whether this is documented anywhere, but I think it's a reasonable assumption and something that people are likely to be relying upon, even though they may not realise it.

It's reasonable, but I also think one should never have this kind of routes (conflicting routes) precisely because the behavior can be unpredictable. So, I intentionally never documented it. The current sorting order is indeed stable, but I would say one should not rely on this so I feel reluctant to add such test cases.

I think there are safer paths to not change the current matcher behavior. I should post these thoughts on the other PRs

posva:
Let's not mark this behavior as intended, at least yet. I think it gives more room for changes in the future

PR #2144, opened 2024-02-19
chore: update chromedriver to 121.0.2 to fix binary CDN URL issue

haoqunjiang:
The default CDN URL for Chrome binaries has been updated in the past few days. So fresh installations of the chromedriver package will fail with a 404 error. Here's an example in `vuejs/ecosystem-ci`: https://github.com/vuejs/ecosystem-ci/actions/runs/7954792731/job/21714823187#step:7:5745

This issue was fixed in `chromedriver` 121.0.2:
https://github.com/giggio/node-chromedriver/commit/06656e982bf7e2ab4ff18ca73dc70ec3032765c0

### DIFF
--- a/packages/router/package.json
+++ b/packages/router/package.json
@@ -133,7 +133,7 @@
     "@vue/server-renderer": "^3.3.11",
     "@vue/test-utils": "^2.4.3",
     "browserstack-local": "^1.5.5",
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #2154, opened 2024-03-01
fix: combining 'end' and 'strict'

skirtles-code:
This Playground illustrates the problem:

- [Playground](https://play.vuejs.org/#eNp9U8Fu2zAM/RVBFzdAbK1It4PhBtmGAtuAbkW3W7WDYzOxW1sSJDnN4PrfR0lN7G5pfTH1+Eg+klJPPyqV7DqgKc1MoWtliQHbqSUXdauktqQnhYbcwjW0Uv/5UhuLv/kzeCs7C5oMZKNlSyJMFGsPRVxwUUhhLAkAuXwRctZzQUgVsqWnSpzN5o7io01K7tyBEB/mPpG3kJLIu88jT3Wfym2FMOsMaDbCxuq6sCmxuoMjCKJMySZvzAgVEpsWIJDaDwH0v99cDLOxJVfFYEd3oVA0f66IQsbaB+s8wuiMheniXPFgoVUNNownQjKbrxsga6lL0B5x2MFydrW8wYIZQ2MKXue2qMBMcDQnGcgu3kh9yamTS2oRZHM6SVIuM6Vh2ffeR4YhY+6MecqTrLDLRIORzQ7OXNAscasgT08kiqNTGY6i0HCdoo3WYQJ0Tq3BqW7qbXJvpMCL6HfMqdtF3YD+oWyNU+cUdxIycpo3jXz85rHJSjGmguLhBH5v9g7j9Aalg94Bp0efzfUWbHBf/fwOe7SPzlaWXYPsN5y3bhid0xhonzpRouwJz6v96p9TLba/zNXegjCHppxQxxw8n1N8RJ/faH2Uu0gWPg7vJk4xPNe4zdU/cwyOl0nGhxo0V9YqkzLWCfWwTXDybGSsLpJF8o6V+CwnaAKmjddaPuI4seCk2xWSWAk7K2Vj4lzVr5X4j7j6kFwk71lTrxlmZ7UoYe9zH9sc/gJqEYgy)

The bottom row is the one that isn't working correctly. This combination of options:

```js
{
  path: '/user/',
  strict: true,
  end: false,
  component: ...
}
```

should match the location path `/user/1`, but it doesn't.

The reason why is the RegExp is checking for the `/` twice:

```
^/user/(?:/|$)
```

---

While I've added a lot of extra testing for `end: false`, only two of those assertions would have failed with the previous code. Specifically, this one on line 666:

```js
matchParams('/home/', '/home/other', {}, options)
```

and this one on line 673:

```js
matchParams('/home/:p/', '/home/a/b', { p: 'a' }, options)
```

In both cases they would previously have failed to resolve the route.

### DIFF
--- a/packages/router/__tests__/matcher/pathParser.spec.ts
+++ b/packages/router/__tests__/matcher/pathParser.spec.ts
@@ -619,8 +619,59 @@ describe('Path parser', () => {
     })
 
     it('can not match the end', () => {
-      matchParams('/home', '/home/other', null, { end: true })
-      matchParams('/home', '/home/other', {}, { end: false })
+      const options = { end: false }
+
+      matchParams('/home', '/home', {}, options)
+      matchParams('/home', '/home/', {}, options)
+      matchParams('/home', '/home/other', {}, options)
+      matchParams('/home', '/homepage', {}, options)
+
+      matchParams('/home/:p', '/home', null, options)
+      matchParams('/home/:p', '/home/', null, options)
+      matchParams('/home/:p', '/home/a', { p: 'a' }, options)
+      matchParams('/home/:p', '/home/a/', { p: 'a' }, options)
+      matchParams('/home/:p', '/home/a/b', { p: 'a' }, options)
+      matchParams('/home/:p', '/homepage', null, options)
+
+      matchParams('/home/', '/home', {}, options)
+      matchParams('/home/', '/home/', {}, options)
+      matchParams('/home/', '/home/other', {}, options)
+      matchParams('/home/', '/homepage', {}, options)
+
+      matchParams('/home/:p/', '/home', null, options)
+      matchParams('/home/:p/', '/home/', null, options)
+      matchParams('/home/:p/', '/home/a', { p: 'a' }, options)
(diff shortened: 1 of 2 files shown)


packages/router/__tests__/matcher/pathParser.spec.ts
```diff
@@ -619,8 +619,59 @@ describe('Path parser', () => {
     })
 
     it('can not match the end', () => {
-      matchParams('/home', '/home/other', null, { end: true })
```
posva, line 622: Was this test removed by accident? It would be great to have it

posva (commented):
I'm still trying to figure out this one. I have never been a fan of the `end` option. I may not have included enough tests on how this should work. I will have to check them again

packages/router/__tests__/matcher/pathParser.spec.ts
skirtles-code: It was intentional, but I've just pushed another commit to bring it back. This also adds extra tests to check that `end: true` works as expected in various simple cases, either when passed explicitly or when left as the default.

PR #2157, opened 2024-03-02
feat(dx): warn when `addRoute` cannot find the parent

skirtles-code:
This adds a warning when calling `router.addRoute('foo', { ... })` if the parent route isn't found.

Without this warning, it can be difficult to tell why `addRoute` isn't working correctly. With an absolute path, beginning with `/`, the addition seems to succeed, but the new route won't be a child of the intended parent. With a relative path, it'll fail to add the new route, but with a confusing error about the `path` being invalid.

I've also added a test for the similar warning in `removeRoute`, though the warning itself already existed.

### DIFF
--- a/packages/router/__tests__/router.spec.ts
+++ b/packages/router/__tests__/router.spec.ts
@@ -1076,5 +1076,22 @@ describe('Router', () => {
         name: 'Param',
       })
     })
+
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

Issue #992, opened 2021-06-09
Uncaught TypeError: Cannot use 'in' operator to search for 'path' in undefined


1091287322:
### Version
4.0.8

### Reproduction link
[https://jsfiddle.net/1091287322/6rj3d9gw/4/](https://jsfiddle.net/1091287322/6rj3d9gw/4/)

### Steps to reproduce
很简单的应用，下面是我的代码:
// router.js
const Home = { template: '<div>Home</div>' }
const routes = [
	{ path: '/', component: Home }
]
let router = createRouter({
	history: createWebHistory,
	routes: routes
})
// main.js
const app = createApp(App)
app.use(router)
app.mount('#app')

它可以成功编译，但不能运行
浏览器报了这个错误：

Uncaught TypeError: Cannot use 'in' operator to search for 'path' in undefined
    at resolve (vue-router.esm-bundler.js:2837)
    at pushWithRedirect (vue-router.esm-bundler.js:2942)
    at push (vue-router.esm-bundler.js:2911)
    at Object.install (vue-router.esm-bundler.js:3289)
    at Object.use (runtime-core.esm-bundler.js:4001)
    at Module.Vtdi (main.js:9)
    at __webpack_require__ (bootstrap:789)
    at fn (bootstrap:100)
    at Object.0 (index.js:4)
    at __webpack_require__ (bootstrap:789)

能告诉我如何解决这个问题吗？

### What is expected?
能正常使用

### What is actually happening?
浏览器直接报错，很明显这是4.0.8这个版本最明显的BUG

GGwujun:
createWebHistory should be called as a function. [createrouter](https://next.router.vuejs.org/api/#createrouter)
as follows:

```javascript
let router = createRouter({
  history: createWebHistory(),
  routes: routes
})
```

posva:
Please, next time consider using the [forum](http://forum.vuejs.org/), the [Discord server](https://vue-land.js.org/) or [StackOverflow](http://stackoverflow.com/tags/vue.js) for questions first. But feel free to come back and open an issue if it turns out to be a bug 🙂
Also open issues in English

PR #2158, opened 2024-03-03
feat(dx): warn when passing undefined/null locations

skirtles-code:
This PR attempts to improve the warnings and error handling for a common problem encountered during development.

## Background

There's an error message I've encountered several times on Vue Land:

> TypeError: Cannot use 'in' operator to search for 'path' in undefined
 
This comes from `router.resolve()`, e.g.:

- [Playground](https://play.vuejs.org/#eNp9UV1PgzAU/SsNL2zJgESnDyQmi2aJmviR6Zv1gcEddIOWtIVhCP/d24L70DleWs49Pfece1uniBj318oJHVaUQmrSklhCpOEJCiG/7pnSeEwGcCEqDZJ0ZCVFQdy6Ak9ayKWc8lhwpUkPkJujJ6OWckKyXi081WI0nhiKfa1C8mF+CLHPzFdGOguJG7iWZb5YoGEOXIek7XrQHp+Ud2NjpzfiS1Air2FU8QRWjEMydiZDWK+ISgwvOMa3nehQUNRB1V6UOvuYBqZOpnWpwiCoeLlJffQR7BmzqX/hXwUJhjpAfVCFt5Riq/C+RvUhBXVmSAoSqLUQufKikv3X4g9xdu1PsVPOlgGqBwzjNVbbSOMkOoypFe5kxdJfIc3oWA7ypdQMd3YUNspzsX20mJYV7IzGGcSbE/haNb3lVxw0yBoOwulIpqD78vztGRq874qFSKoc2WeKC7O6ynjsabe4Q7R9wLNuH+zOGE/f1bzRwNVPKGPUTsPy7SLvzkTf2730p7spdt+TZA3R)

That's a silly example, it's rarely quite so obvious when someone is having this problem in a real application.

These two issues are both hitting the same problem (the first one is in the wrong repo):
- https://github.com/vuejs/vue-router/issues/3735
- https://github.com/vuejs/router/issues/992

You get something similar with `null` instead of `undefined`.

`router.push(undefined)` gives the same error, as that uses `router.resolve()`.

Most commonly this is hit via `useLink()`. Here's a simple example of that:

- [Playground](https://play.vuejs.org/#eNp9Uk2P0zAQ/SujcEgqtYkEhUOkRQW0EiC+tHAjHLLJtHGb2JY97hZV/e+M7fRjl931Je7M85v33nSfDLWQ+domZSIGrQzBHhqDNeE7reEAS6MGSLcO00o+AHzFQZm/H4Ul/kzH4o1yhGYKzuIXITeXDDMTekxUyUZJSxALcHXvbbavJEAXacvHZmWTqYeE17aE3/4HQHjmj66pKyEt0oDyp1GsXKKkEvaHWAyfP5U8TM5yanZ81MLuoxCL5HQ2OdKPvmLPH1IlONniUkhsR+7A6W+GHxsJ/PzqLaQd9r1i+2F6BPHInCmzmMSpNCgnKUtf8D2dJNNxNbOh1rwqJXlZYX41NmyVsLM4skrOUftylXRE2pZF4aTerHLOojgjFvP8Zf66aDnYi2qOdpjdGnVn+b5m9jHJKlkwqGhxS0r1dlZr8dSI/4CLN/mcJ/XitmD2QnBku8B9zINtkuVFLMXqgUm/PtGj+a5J8KLuma050rvPoUbG4Ulo02GzeaS+trso+YdBdrfFC3NUmxVSbF///IY7vp+ag2pdz+hnmjdoVe+8xgh7z/8Lln2BC2o/hZ0Jufplr3eE0h5NeaEhjYAPi/zwjPWz3Ff5/JTi4R/NvkZT)

Again, I've made it obvious what the problem is with `to: undefined`, but real code is rarely so blatant.

It is possible to hit this same error via `RouterLink`, but in that case we get a prop validation warning, making it a bit easier to figure out what the problem is:

- [Playground](https://play.vuejs.org/#eNqFVNuO0zAQ/ZUhPCSV2ljAwkPURbuglRbETQuCB8JDtnFb7ya2ZTvdoir/zviS21LKQxt75oznjOeMD1FdMJ7e6SiLWC2FMnCAlaKFoZdSQgtrJWqIdw2Ncx4ASjSGquBJid8ObhvX+QopUx+b8/7UBH+znAOkjaaJD/f7WjTcJPFTDItn0TwK4chtqVeKSQOamka+7nMd4MaFf2D8fh7W3xl9GBNf9ASRg+DaQIVoDefw0yY92D8AQ/cmg/ha1DSee5MszBZNBCMB2vkj5OUtHjuCcgstnNHhc/4r50vieSNj3BhaywrvAHcAS17s3AKXnQN2i7VQ53mUWIpzYLyk+xl+POU8CgEYomXBEc7WiHYwdMICMCE6BthwPZAZgVh7UGorQ/zh4M5NbUXQtksyoDtmZMqZdKSXVjYdaHTvJOCCexSP7fSNOCq2j7QW6vc10wY/82D05x7vZYi3/XKJO8V1hiC7TpO2LxNcb+n1SfcOWtJ10VRmQiFxrd96dtkxysnMKcHR05lXVq+tkZSCXgBWAplxylFKHWXvGoQ2hHlZHYvtywjB9oO6a+30jK/i1Agpun406N2g1HqDY4L+JL6mVSXgh1BV+STGaT2l7O0zKy0bbEWFO2dlXDYGJVuLklaoRPQ7QU81MunLEdaE/De3vwp3EDTa69GZp5l8/Yu6kKhIwTGXu3ccJ+fAacu6TuTRoD1rzqOtMVJnhDRc3m9SbAgZEBdn6fP0JSlRGSNrSnW9uFXiQbsRyKPQzjy6QBAp6c4IUelFIdm/UvwFvHiVnmGmit0SPJ24d8CdHV6gFss0Gnu5ZptHRVoNsYqqz9Iw7PWk2AJb/fDe2YxqaE90taWr+yP2O42vj6X8RVGsbkdHxZlCbajx7quvn/ClGTlRCk2F6BPOG6pF1ViOHvam4SXSHuEc23euZ4xvvumrvaFcd0VZov799nXjFb49UfpA90V61t9i+wfNoGNa)

The mistake in that example is a typo in `patn: '/about'`.

But even with the prop validation warning, it can be difficult to identify exactly which component has the problem. If we could inspect the components using Vue Devtools it'd be much easier, but that isn't currently an option because rendering bombs out.

## The Vuetify problem

Someone on Vue Land was hitting this error, but without any warnings to indicate what they were doing wrong. They weren't using `RouterLink`, they were using Vuetify's breadcrumb component.

The ultimate cause of the problem was an `undefined` in their data that they hadn't taken into account. But this isn't quite as simple as it sounds.

There's this problematic line in Vuetify:

- https://github.com/vuetifyjs/vuetify/blob/c8503b1e0d9669bfc384ba563d5211922ea33e80/packages/vuetify/src/composables/router.tsx#L68

The problem here is that `props.to` can be `undefined`. While the Vuetify composable does check that, it only checks the initial value. In the application we encountered on Vue Land, the `to` prop was populated initially, only becoming `undefined` after some new data loaded.

It isn't obvious to me how Vuetify could fix this problem in their own code. There doesn't seem to be a way to disable the `useLink()` composable when the prop is `undefined` and re-enable it when the value is defined.

This PR doesn't attempt to address that problem directly, it just aims to improve the warnings/errors to ease debugging.

## What's changed

- `router.resolve()` no longer throws an error in dev. It logs a warning instead. This allows Vue Devtools to be used to debug the problem.
- `useLink()` logs warnings. This is usually a bit more helpful than the warning in `router.resolve()`, as it has more context to work with.
- Vue Devtools will show an error label if `props.to` is invalid, making it much easier to find a rogue `RouterLink` in a long list:
  Vue Devtools

In cases where `RouterLink` has an `undefined` or `null` value for `to`, this will lead to 3 warnings being shown: one for prop validation, one for `useLink()` and one for `router.resolve()`. I couldn't see an easy way to avoid that, but I think that's much better than not having the warnings.

### DIFF
--- a/packages/router/__tests__/router.spec.ts
+++ b/packages/router/__tests__/router.spec.ts
@@ -330,7 +330,7 @@ describe('Router', () => {
     expect(route1.params).toEqual({ p: 'a' })
   })
 
-  it('handles an undefined location', async () => {
+  it('warns on undefined location during dev', async () => {
     const { router } = await newRouter()
 
     const route1 = router.resolve(undefined as any)
@@ -338,7 +338,7 @@ describe('Router', () => {
     expect(route1.path).toBe('/')
   })
 
-  it('handles a null location', async () => {
+  it('warns on null location during dev', async () => {
     const { router } = await newRouter()
 
     const route1 = router.resolve(null as any)
--- a/packages/router/__tests__/router.spec.ts
+++ b/packages/router/__tests__/router.spec.ts
@@ -330,6 +330,22 @@ describe('Router', () => {
     expect(route1.params).toEqual({ p: 'a' })
   })
 
+  it('handles an undefined location', async () => {
+    const { router } = await newRouter()
+
+    const route1 = router.resolve(undefined as any)
+    expect('router.resolve() was passed an invalid location').toHaveBeenWarned()
+    expect(route1.path).toBe('/')
+  })
+
+  it('handles a null location', async () => {
+    const { router } = await newRouter()
+
+    const route1 = router.resolve(null as any)
+    expect('router.resolve() was passed an invalid location').toHaveBeenWarned()
+    expect(route1.path).toBe('/')
+  })
+
   it('removes null/undefined optional params when current location has it', async () => {
     const { router } = await newRouter()
 
--- a/packages/router/__tests__/useLink.spec.ts
+++ b/packages/router/__tests__/useLink.spec.ts
@@ -7,9 +7,7 @@ import { mockWarn } from 'jest-mock-warn'
 import {
   createMemoryHistory,
   createRouter,
-  routeLocationKey,
   RouteLocationRaw,
-  routerKey,
   useLink,
   UseLinkOptions,
 } from '../src'
@@ -50,10 +48,7 @@ async function callUseLink(args: UseLinkOptions) {
     },
     {
       global: {
-        provide: {
-          [routerKey as any]: router,
-          [routeLocationKey as any]: router.currentRoute.value,
-        },
+        plugins: [router],
       },
     }
   )
--- a/packages/router/__tests__/useLink.spec.ts
+++ b/packages/router/__tests__/useLink.spec.ts
@@ -1,0 +1,148 @@
+/**
+ * @jest-environment jsdom
+ */
+import { nextTick, ref } from 'vue'
+import { mount } from '@vue/test-utils'
+import { mockWarn } from 'jest-mock-warn'
+import {
+  createMemoryHistory,
+  createRouter,
+  routeLocationKey,
+  RouteLocationRaw,
+  routerKey,
+  useLink,
+  UseLinkOptions,
+} from '../src'
+
+async function callUseLink(args: UseLinkOptions) {
+  const router = createRouter({
+    history: createMemoryHistory(),
+    routes: [
+      {
+        path: '/',
+        component: {},
+        name: 'root',
+      },
+      {
+        path: '/a',
+        component: {},
+        name: 'a',
+      },
+      {
+        path: '/b',
+        component: {},
+        name: 'b',
+      },
+    ],
+  })
+
+  await router.push('/')
+
+  let link: ReturnType<typeof useLink>
+
+  mount(
+    {
+      setup() {
+        link = useLink(args)
+
+        return () => ''
+      },
+    },
+    {
+      global: {
+        provide: {
+          [routerKey as any]: router,
+          [routeLocationKey as any]: router.currentRoute.value,
+        },
+      },
+    }
+  )
+
+  return link!
+}
+
+describe('useLink', () => {
+  describe('basic usage', () => {
+    it('supports a string for "to"', async () => {
+      const { href, route } = await callUseLink({
+        to: '/a',
+      })
+
+      expect(href.value).toBe('/a')
+      expect(route.value).toMatchObject({ name: 'a' })
+    })
+
+    it('supports an object for "to"', async () => {
+      const { href, route } = await callUseLink({
+        to: { path: '/a' },
+      })
+
+      expect(href.value).toBe('/a')
+      expect(route.value).toMatchObject({ name: 'a' })
+    })
+
+    it('supports a ref for "to"', async () => {
+      const to = ref<RouteLocationRaw>('/a')
+
+      const { href, route } = await callUseLink({
+        to,
+      })
+
+      expect(href.value).toBe('/a')
+      expect(route.value).toMatchObject({ name: 'a' })
+
+      to.value = { path: '/b' }
+
+      await nextTick()
+
+      expect(href.value).toBe('/b')
+      expect(route.value).toMatchObject({ name: 'b' })
+    })
+  })
+
+  describe('warnings', () => {
+    mockWarn()
+
+    it('should warn when "to" is undefined', async () => {
+      await callUseLink({
+        to: undefined as any,
+      })
+
+      expect('Invalid value for prop "to" in useLink()').toHaveBeenWarned()
+      expect(
+        'router.resolve() was passed an invalid location'
+      ).toHaveBeenWarned()
+    })
+
+    it('should warn when "to" is an undefined ref', async () => {
+      await callUseLink({
+        to: ref(undefined as any),
+      })
+
+      expect('Invalid value for prop "to" in useLink()').toHaveBeenWarned()
+      expect(
+        'router.resolve() was passed an invalid location'
+      ).toHaveBeenWarned()
+    })
+
+    it('should warn when "to" changes to a null ref', async () => {
+      const to = ref('/a')
+
+      const { href, route } = await callUseLink({
+        to,
+      })
+
+      expect(href.value).toBe('/a')
+      expect(route.value).toMatchObject({ name: 'a' })
+
+      to.value = null as any
+
+      await nextTick()
+
+      expect('Invalid value for prop "to" in useLink()').toHaveBeenWarned()
+      expect(
+        'router.resolve() was passed an invalid location'
+      ).toHaveBeenWarned()
+    })
+  })
+})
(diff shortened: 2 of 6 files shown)


packages/router/src/RouterLink.ts
```diff
@@ -93,7 +95,40 @@ export function useLink(props: UseLinkOptions) {
   const router = inject(routerKey)!
   const currentRoute = inject(routeLocationKey)!
 
-  const route = computed(() => router.resolve(unref(props.to)))
+  const isValidTo = (to: unknown) => typeof to === 'string' || isObject(to)
+  let hasPrevious = false
+  let previousTo: unknown = null
```
posva, line 100: I think these can be merged in one

packages/router/src/RouterLink.ts
```diff
@@ -83,6 +84,7 @@ export interface UseLinkDevtoolsContext {
   route: RouteLocationNormalized & { href: string }
   isActive: boolean
   isExactActive: boolean
+  error: string
```
posva, line 87: Small nitpick: `string | null` seems more appropriate

packages/router/src/RouterLink.ts
```diff
@@ -93,7 +95,40 @@ export function useLink(props: UseLinkOptions) {
   const router = inject(routerKey)!
   const currentRoute = inject(routeLocationKey)!
 
-  const route = computed(() => router.resolve(unref(props.to)))
+  const isValidTo = (to: unknown) => typeof to === 'string' || isObject(to)
```
posva, line 98: Let's inline this and `isObject` checks

packages/router/src/devtools.ts
```diff
@@ -119,10 +119,16 @@ export function addDevtools(app: App, router: Router, matcher: RouterMatcher) {
...
             let backgroundColor = ORANGE_400
             let tooltip: string = ''
+            let textColor = 0
 
-            if (devtoolsData.isExactActive) {
+            if (devtoolsData.error) {
+              label = devtoolsData.error
+              backgroundColor = RED_100
+              textColor = RED_700
+            } else if (devtoolsData.isExactActive) {
```
posva, line 131: Nice!

packages/router/src/utils/index.ts
```diff
@@ -37,3 +37,6 @@ export const noop = () => {}
  */
 export const isArray: (arg: ArrayLike<any> | any) => arg is ReadonlyArray<any> =
   Array.isArray
+
+export const isObject = (val: unknown): val is Record<any, any> =>
+  val !== null && typeof val === 'object'
```
posva, line 42: Usually I write this one
```suggestion
  val && typeof val === 'object'
```

I'm curious if you prefer this version for any particular reason

packages/router/__tests__/router.spec.ts
```diff
@@ -330,6 +330,22 @@ describe('Router', () => {
     expect(route1.params).toEqual({ p: 'a' })
   })
 
+  it('handles an undefined location', async () => {
```
posva, line 333:
```suggestion
  it('warns on undefined location during dev', async () => {
```

packages/router/__tests__/router.spec.ts
```diff
@@ -330,6 +330,22 @@ describe('Router', () => {
...
+    const { router } = await newRouter()
+
+    const route1 = router.resolve(undefined as any)
+    expect('router.resolve() was passed an invalid location').toHaveBeenWarned()
+    expect(route1.path).toBe('/')
+  })
+
+  it('handles a null location', async () => {
```
posva, line 341:
```suggestion
  it('warns on null location during dev', async () => {
```

packages/router/__tests__/useLink.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+      },
+    },
+    {
+      global: {
+        provide: {
+          [routerKey as any]: router,
+          [routeLocationKey as any]: router.currentRoute.value,
+        },
+      },
+    }
```
posva, line 58: was passing the `router` instance as a plugin failing?

posva (changes requested):
I think this will be really useful! Thank you

packages/router/src/RouterLink.ts
skirtles-code: I assume you intended for `null` to be used instead of `''` when there's no error. I've updated accordingly.

packages/router/src/utils/index.ts
skirtles-code: I'd usually write it the way you suggested.

I wasn't sure about the TS types for that function so I just copied the equivalent function in Vue core:

- https://github.com/vuejs/core/blob/f66a75ea75c8aece065b61e2126b4c5b2338aa6e/packages/shared/src/general.ts#L52

I'm happy either way.

packages/router/src/RouterLink.ts
skirtles-code: I tried that when I first wrote it, but it didn't work out.

The problem is picking an initial value that can be used as a sentinel. `null` and `undefined` don't work, because those could be set later.

I tried using a Symbol or object instead, something like this:

```js
const INITIAL_VALUE = {}
let previousTo: unknown = INITIAL_VALUE
```

with subsequent checks for `previousTo === INITIAL_VALUE`, but that seemed more complicated than a separate `hasPrevious` flag.

packages/router/src/RouterLink.ts
skirtles-code: I can make this change, but I want to make sure we're on the same page before I do.

If I've understood correctly, you're proposing removing the `isValidTo` and `isObject` functions and just writing the code inline instead?

- The `isValidTo` function is used in two places, so I'd have to inline it twice.
- `isObject` is used in two places, but if I inline `isValidTo` too then I'd need to inline `isObject` in three places.

`isObject` could potentially be used in various other places. That seemed like a separate refactoring so I didn't do it in this PR, but we have various places with `typeof` checks for `object`.

I've also just noticed the `isRouteLocation` function, which seems to be doing that same thing as `isValidTo`:

- https://github.com/vuejs/router/blob/63770832d01a86a7464fde744e94058aae07de2e/packages/router/src/types/typeGuards.ts#L3

Should I maybe be using that instead?

posva: You are right, using `isRouteLocation()` is even better.

packages/router/src/utils/index.ts
posva: I guess you could just use isRouteLocation where isObject is used in router.ts

posva (approved):
Thanks!

PR #2160, opened 2024-03-05
docs(zh): translate the warning part of passing_props to Chinese

make17better:
### Finished the translation of Passing Props to Route Components

Before: 
图片

After: 
图片


make17better:
Sorry, there's something wrong, I have submitted a new PR #2161 to fix it.

PR #2173, opened 2024-03-08
docs: fix typo

make17better:
### Fix typo

**Description:** A period is missing between `instances` and `Having`, click [here](https://router.vuejs.org/zh/api/interfaces/RouteRecordNormalized.html#Properties-instances) for details.

**Snapshot:**
图片

### DIFF
--- a/packages/docs/api/interfaces/RouteRecordNormalized.md
+++ b/packages/docs/api/interfaces/RouteRecordNormalized.md
@@ -53,7 +53,7 @@ ___
 
 • **instances**: `Record`\<`string`, `undefined` \| ``null`` \| `ComponentPublicInstance`\>
 
-Mounted route component instances
+Mounted route component instances.
 Having the instances on the record mean beforeRouteUpdate and
 beforeRouteLeave guards can only be invoked with the latest mounted app
 instance if there are multiple application instances rendering the same


make17better:
I've already added a period, but why it still doesn't work in the [Deploy Preview](https://router.vuejs.org/api/interfaces/RouteRecordNormalized.html#instances)? Is this caused by the VitePress configuration or the Markdown syntax of VitePress?

posva (changes requested):
Can you fix it in the source code please? This is generated

make17better:
> Can you fix it in the source code please? This is generated

OK, I got it, I will fix it.

make17better:
> Can you fix it in the source code please? This is generated

Actually, I've changed it in the source code before I open this PR. Seems to be a Netlify deployment issue? Because everything worked fine in the dist got after I executed `pnpm run docs:build` locally. Finally, I executed `pnpm run docs:preview`, it also works.

I'm sorry I can't fix this problem because I can't (or I just don't know how to) debug in a cloud service environment like the Netlify Deploy. Please feel free to enlighten me.

Here's the screenshot of my local preview: 
图片

posva:
I see, thank you. In that case it will appear in the next deployment

PR #2174, opened 2024-03-09
docs: update examples to use SFCs

skirtles-code:
## Objectives

- Switch examples to use SFC syntax.
- Include Composition API examples with `<script setup>`.
- Update the wording to take account of the Composition API, including changing any wording that makes it sound like a new, exotic choice.

There are lots of other changes that might be desirable, but I've tried to stick to that scope as much as possible. For the most part, the order and flow of the existing docs has been retained.

These changes need making in one big jump to keep the guide consistent. Migrating one page at a time would just make the guide highly confusing.

I am aware that big changes like these are a burden for translators. I've tried to take that into account where possible.

## General changes

In many cases, the change just involved switching an existing example to use SFC syntax.

In examples that needed a `<script>` section, both Composition API and Options API examples have been provided. This is using a VitePress [code group](https://vitepress.dev/guide/markdown#code-groups). The source `.md` file ends up looking much longer, but the code group keeps the examples relatively brief in the generated docs.

Mentions of `$router` or `$route` have been updated, where necessary. In the templates those are retained, but in the main text they are generally replaced with `router` and `route` respectively. In cases where it seems particularly necessary there is further explanation about what those variables are, though it is already covered in the `Getting Started` page at the beginning of the guide.

In parts of the docs that already mentioned the Composition API, it tended to be described as 'new', or presented as an exotic or alien alternative to the Options API. Those sections also tended to predate `<script setup>` and referenced the use of a `setup` function. The wording of those sections has been updated to present the use of `<script setup>` as a normal way to use Vue.

### `composition-api.md` - Composition API

This is a bit of a tricky page to update. I don't want to remove the page entirely, as that would need bigger changes elsewhere, but having a separate page for the Composition API does feel a bit unnatural when it's now being used in most of the other pages.

### `data-fetching.md` - Data Fetching

I'm not sure what to do about the examples on this page. The first one was easily updated, but it isn't clear to me how useful that example is in isolation. It doesn't really demonstrate much from a Vue Router perspective.

The second example is much more enlightening, but it only works with the Options API. It could be written using the Composition API and `defineOptions`, but as discussed in #2131, that isn't necessarily the way to go in a docs example. But I don't know what to show instead. A future plan based on data loaders doesn't really help people reading the docs today.

Perhaps that second section should include a warning box that explicitly notes the lack of a Composition API example? I think that would be better than the example being conspicuously absent without an explanation.

### `installation.md` - Installation

The diff on GitHub is a bit difficult to follow. All I've done is swap the order of the sections. This brings it into line with the rest of the docs, which now assume build tools and SFCs.

### `index.md` - Getting Started

This is probably the most important change in this PR.

This page has been almost entirely rewritten, though the overall objective and approach is similar to the old page.

I think the `Getting Started` page is primarily aimed at beginners. Experienced users are more likely to focus on the later pages, looking for details about the specific features they're using. While we'd obviously like all pages to be understandable by beginners, I think it makes sense for the `Getting Started` page to be a little more explicit in its explanations and take things a bit slower.

There are a few things I've done with that in mind:
- Briefly explaining what Vue Router is and what it's for.
- Explicitly stating the prerequisites near the top.
- Allowing key ideas to be repeated if they seem relevant.
- Explaining core Vue concepts that beginners may have overlooked or forgotten. In general, this is not the job of the Router docs, but I think it's worth it here.
- Establishing the conventions that will be used in the rest of the guide.

Some parts of that were inspired by this feedback about the current page: https://github.com/vuejs/router/discussions/2068#discussioncomment-8634606.

### DIFF
--- a/packages/docs/guide/advanced/composition-api.md
+++ b/packages/docs/guide/advanced/composition-api.md
@@ -2,90 +2,84 @@
 
 <VueSchoolLink
   href="https://vueschool.io/lessons/router-and-the-composition-api"
-  title="Learn how to use Vue Router with the composition API"
+  title="Learn how to use Vue Router with the Composition API"
 />
 
-The introduction of `setup` and Vue's [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html), open up new possibilities but to be able to get the full potential out of Vue Router, we will need to use a few new functions to replace access to `this` and in-component navigation guards.
+The introduction of Vue's [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html) opened up new possibilities, but to be able to get the full potential out of Vue Router, we will need to use a few new functions to replace access to `this` and in-component navigation guards.
 
 ## Accessing the Router and current Route inside `setup`
 
-Because we don't have access to `this` inside of `setup`, we cannot directly access `this.$router` or `this.$route` anymore. Instead we use the `useRouter` and `useRoute` functions:
+Because we don't have access to `this` inside of `setup`, we cannot directly access `this.$router` or `this.$route`. Instead, we use the `useRouter` and `useRoute` functions:
 
-```js
+```vue
+<script setup>
 import { useRouter, useRoute } from 'vue-router'
 
-export default {
-  setup() {
-    const router = useRouter()
-    const route = useRoute()
+const router = useRouter()
+const route = useRoute()
 
-    function pushWithQuery(query) {
-      router.push({
-        name: 'search',
-        query: {
-          ...route.query,
-          ...query,
-        },
-      })
-    }
-  },
+function pushWithQuery(query) {
+  router.push({
+    name: 'search',
+    query: {
+      ...route.query,
+      ...query,
+    },
+  })
 }
+</script>
 ```
 
-The `route` object is a reactive object, so any of its properties can be watched and you should **avoid watching the whole `route`** object. In most scenarios, you should directly watch the param you are expecting to change
+The `route` object is a reactive object. In most scenarios, you should **avoid watching the whole `route`** object. Instead, you can directly watch the properties you are expecting to change:
 
-```js
+```vue
+<script setup>
 import { useRoute } from 'vue-router'
 import { ref, watch } from 'vue'
 
-export default {
-  setup() {
-    const route = useRoute()
-    const userData = ref()
+const route = useRoute()
+const userData = ref()
 
-    // fetch the user information when params change
-    watch(
-      () => route.params.id,
-      async newId => {
-        userData.value = await fetchUser(newId)
-      }
-    )
-  },
-}
+// fetch the user information when params change
+watch(
+  () => route.params.id,
+  async newId => {
+    userData.value = await fetchUser(newId)
+  }
+)
+</script>
 ```
 
-Note we still have access to `$router` and `$route` in templates, so there is no need to return `router` or `route` inside of `setup`.
+Note we still have access to `$router` and `$route` in templates, so there's no need to use `useRouter` or `useRoute` if we only need those object in the template.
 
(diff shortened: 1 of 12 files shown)


packages/docs/guide/advanced/composition-api.md
```diff
@@ -2,90 +2,84 @@
...
+  title="Learn how to use Vue Router with the Composition API"
 />
 
-The introduction of `setup` and Vue's [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html), open up new possibilities but to be able to get the full potential out of Vue Router, we will need to use a few new functions to replace access to `this` and in-component navigation guards.
+The introduction of Vue's [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html) opened up new possibilities, but to be able to get the full potential out of Vue Router, we will need to use a few new functions to replace access to `this` and in-component navigation guards.
 
 ## Accessing the Router and current Route inside `setup`
 
-Because we don't have access to `this` inside of `setup`, we cannot directly access `this.$router` or `this.$route` anymore. Instead we use the `useRouter` and `useRoute` functions:
+Because we don't have access to `this` inside of `setup`, we cannot directly access `this.$router` or `this.$route`. Instead, we use the `useRouter` and `useRoute` functions:
```
posva, line 12:
```suggestion
Because we don't have access to `this` inside of `setup`, we cannot directly access `this.$router` or `this.$route`. Instead, we use the `useRouter` and `useRoute` composables:
```

packages/docs/guide/advanced/data-fetching.md
```diff
@@ -27,9 +29,54 @@ Let's assume we have a `Post` component that needs to fetch the data for a post
...
+const route = useRoute()
+
+const loading = ref(false)
+const post = ref(null)
+const error = ref(null)
+
+// watch the params of the route to fetch the data again
+watch(() => route.params.id, fetchData, { immediate: true })
+
+async function fetchData() {
```
posva, line 47: let's add the id as a paremeter rather than relying on the route instance within the function

packages/docs/guide/advanced/data-fetching.md
```diff
@@ -41,7 +88,7 @@ export default {
   created() {
     // watch the params of the route to fetch the data again
     this.$watch(
-      () => this.$route.params,
+      () => this.$route.params.id,
       () => {
         this.fetchData()
```
posva, line 93: same here, let's pass the parameter too

packages/docs/guide/advanced/extending-router-link.md
```diff
@@ -9,7 +9,51 @@ The RouterLink component exposes enough `props` to suffice most basic applicatio
...
+      :href="href"
+      @click="navigate"
+      :class="isActive ? activeClass : inactiveClass"
+    >
+      <slot />
+    </a>
+  </router-link>
+</template>
+
+<script setup>
```
posva, line 36: let's move setup above template

packages/docs/guide/essentials/dynamic-matching.md
```diff
@@ -57,11 +58,26 @@ A working demo of this example can be found [here](https://codesandbox.io/s/rout
...
+::: code-group
+
+```vue [Composition API]
+<script setup>
+import { watch } from 'vue'
+import { useRoute } from 'vue-router'
+
+const route = useRoute()
+
+watch(() => route.params, (newParams, previousParams) => {
```
posva, line 72: Let's only watch `route.params.id`. Otherwise it contradicts what we say in the introduction

posva (changes requested):
Thanks a lot for this! I took the occasion to comment on some small improvements over the existing code

posva (approved):
Thanks a lot for the great work!

PR #2176, opened 2024-03-16
docs: rewrite the Named Routes page

skirtles-code:
There are several problems I was trying to address:

1. Named routes are used in earlier pages, particularly the **Nested Routes** and **Programmatic Navigation** pages, which use them quite heavily.
2. The list of advantages comes too soon. A reader will struggle to understand that list until they have seen the example.
3. Advantage 2 on the list mentions `encoding/decoding`. As far as I can tell, that should just be `encoding`, as named routes don't impact decoding.
4. Advantage 4 stops mid-sentence. `Bypassing path ranking (e.g. to display a )`.
5. The linked example is for Vue 2.

While I've reworded and reordered much of the page, the material covered should be the same.

I've tweaked the example to use `name: 'profile'`, rather than `name: 'user'`. This was to avoid the `name` coinciding with the `/user/` prefix on the `path`. These kinds of naming coincidences can (in general) lead to beginners misunderstanding what is significant and what isn't, and it seemed unnecessary to risk any confusion here.

### DIFF
--- a/packages/docs/guide/essentials/named-routes.md
+++ b/packages/docs/guide/essentials/named-routes.md
@@ -5,39 +5,41 @@
   title="Learn about the named routes"
 />
 
-Alongside the `path`, you can provide a `name` to any route. This has the following advantages:
-
-- No hardcoded URLs
-- Automatic encoding/decoding of `params`
-- Prevents you from having a typo in the url
-- Bypassing path ranking (e.g. to display a )
+When creating a route, we can optionally give the route a `name`:
 
 ```js
 const routes = [
   {
     path: '/user/:username',
-    name: 'user',
+    name: 'profile', // [!code highlight]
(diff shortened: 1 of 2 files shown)


packages/docs/guide/essentials/named-routes.md
```diff
@@ -5,39 +5,41 @@
...
+
+Using a `name` has various advantages:
 
-In both cases, the router will navigate to the path `/user/erina`.
+- No hardcoded URLs.
+- Automatic encoding of `params`.
+- Prevents you from having a typo in the url.
```
posva, line 36: Let's match the other sentences' style

```suggestion
- Avoids URL typos.
```

posva (approved):
Nice! Thanks a lot

PR #2183, opened 2024-03-21
docs: add 'Active links'

skirtles-code:
This adds a page explaining the RouterLink classes.

This currently isn't covered in the main guide. It is mentioned in the [migration guide for the `exact` prop](https://router.vuejs.org/guide/migration/#Removal-of-the-exact-prop-in-router-link-). There's also an explanation in the [related RFC](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0028-router-active-link.md).

I've put it near the end of the **Essentials** section. It mentions `alias` and `redirect`, so it needed to come after those are introduced. Putting it after the guide to passing props was an arbitrary choice, it could just as easily have gone before.

There's already an advanced guide to **Extending RouterLink**. That page seems to assume that people are already familiar with the existing RouterLink classes and the related concepts of ***active*** and ***exact***. This new page covers those topics.

I haven't covered the edge case around the handling of a `path: ''` on children. I could try to add something about that, though I'm not sure whether it would be more confusing than helpful.

### DIFF
--- a/packages/docs/.vitepress/config/en.ts
+++ b/packages/docs/.vitepress/config/en.ts
@@ -109,8 +109,8 @@ export const enConfig: LocaleSpecificConfig<DefaultTheme.Config> = {
               link: '/guide/essentials/passing-props.html',
             },
             {
-              text: 'RouterLink classes',
-              link: '/guide/essentials/router-link-classes.html',
+              text: 'Active links',
+              link: '/guide/essentials/active-links.html',
             },
             {
               text: 'Different History modes',
--- a/packages/docs/.vitepress/config/en.ts
+++ b/packages/docs/.vitepress/config/en.ts
@@ -109,6 +109,10 @@ export const enConfig: LocaleSpecificConfig<DefaultTheme.Config> = {
               link: '/guide/essentials/passing-props.html',
             },
             {
+              text: 'RouterLink classes',
+              link: '/guide/essentials/router-link-classes.html',
+            },
+            {
               text: 'Different History modes',
               link: '/guide/essentials/history-mode.html',
             },
--- a/packages/docs/guide/essentials/active-links.md
+++ b/packages/docs/guide/essentials/active-links.md
@@ -1,10 +1,10 @@
-# RouterLink classes
+# Active links
 
 It's common for applications to have a navigation component that renders a list of RouterLink components. Within that list, we might want to style links to the currently active route differently from the others.
 
 The RouterLink component adds two CSS classes to active links, `router-link-active` and `router-link-exact-active`. To understand the difference between them, we first need to consider how Vue Router decides that a link is _active_.
 
-## Active links
+## When are links active?
 
 A RouterLink is considered to be ***active*** if:
 
--- a/packages/docs/guide/essentials/router-link-classes.md
+++ b/packages/docs/guide/essentials/router-link-classes.md
@@ -1,0 +1,78 @@
+# RouterLink classes
+
+It's common for applications to have a navigation component that renders a list of RouterLink components. Within that list, we might want to style links to the currently active route differently from the others.
+
+The RouterLink component adds two CSS classes to active links, `router-link-active` and `router-link-exact-active`. To understand the difference between them, we first need to consider how Vue Router decides that a link is _active_.
+
+## Active links
+
+A RouterLink is considered to be ***active*** if:
+
+1. It matches the same route record (i.e. configured route) as the current location.
+2. It has the same values for the `params` as the current location.
+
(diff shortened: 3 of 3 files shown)


posva (commented):
Looks great, thanks again! I'm wondering if we should name the page something else that includes active/matching. We could even omit the RouterLink keyword. So maybe, just "Active Links" works better. What do you think?

posva:
> I haven't covered the edge case around the handling of a path: '' on children. I could try to add something about that, though I'm not sure whether it would be more confusing than helpful.

I think it's fine leaving it out as most people won't have a route that can be visited directly while having a child with path ''.

skirtles-code:
> > I haven't covered the edge case around the handling of a path: '' on children. I could try to add something about that, though I'm not sure whether it would be more confusing than helpful.
> 
> I think it's fine leaving it out as most people won't have a route that can be visited directly while having a child with path ''.

We might be thinking of slightly different edge cases.

To clarify, the specific edge case I'm most concerned about is the one shown here:

- [Playground](https://play.vuejs.org/#eNqlVVtP2zAU/itWNqlFauwNEA9ZQAWEtE27IHZ5WaYpNG5rcOzIdkJRlf++40vSFFr2sEhV7XP5zu3LyToqcybwnY6SiJWVVAat0UzR3NDzqkItmitZolFT01EmgoGStaEqaDDx143a+nW6vKqw981EjzqG30EmEMK1pmPv7u+lrIUZj16B2+ggmkTBHXJL9UyxyiBNTV2dAZgU2iDOxL1Gp+iX9V4jkZc0QaPzEWonW5ILkGTidyZS4nEAAS6GlhWHnOCGUGphETypNkqKxdllrRQVoVxU5WaZgLvXofUavXYKPK85vwYlaiEEeBOPk9a8w+MMNfFcqtMssgkjJnziWRQswObGdeGTVaPEyGAKFhDIObVtSjZGHTThzEcjIVxqx9lpvf1PRh8QCWZBnZJB7XDV5pHbI/bTiG3IOJ8Z1tAkyed23GsLAG030JMEZREae/1BFr3LBNT+xJuuQP9PDGfVQ9j+2kRg9B5qJzE/01Kqx/dMG/ibBKGvdUjYuCdmT80fOrSj42cnCCQdmN1ITvUz2166YbUjogu1YaLvv6MMGhFguRo5QkLxS8YLoFXiLe0TrAcenbF9ek4PZDMJeQrXw66ATtlOduApm/UO0IsXQPtKe2R/gJfIneF/q3gFxQ8nMXZpLP2Qkl2TGx+46L51cGxhB2QizL2q9XLsWwcLxqbvtHTlBlTQeV7zLjKwxQ8uLvMKGCMFcMaFz4ICXrWk60sWwejsPYuWxlQ6IWRW2A1YUM4ahQU1RFQlmYIZUbCPWEnjQpbTI3yID49JAdkP5ZjqMr5V8gFSBZQsCj11cQIHt8PVorpfYOg32VhMj/ERfuPBN9L92C67gjZGSq7jvGL7QjwznJ7gE/wWFsctAXTCREFXDjsMtoV2Gg2TnbPFk2ZaijBO1dfKMJj8VlNzzuXDRyczqqZ9orMlnd3vkN/plU/5WlGoroGR9DqTqwU1Xn317QtdwblXlrKoeRjgHuUN1ZLXNkdvdlGLAtIe2LlsPzhuMLH4rq9WsJJ0V5RN1HXD2btBXr5Q+ibdI3w86OJwuTz7hL30JSpYE1a4hUBNeAtToIdVbC3vEGdrMf1HsK3Xfl/APw1VtlkQByrGh2+j9i/dWsRB)

The routes here are:

```js
const routes = [
  {
    path: '/user',
    children: [
      {
        path: '',
        name: 'A',
        component: UserView
      }, {
        path: 'roles',
        name: 'B',
        component: UserRolesView
      }
    ]
  }
]
```

Route A is a sibling of route B, not an ancestor, yet it is shown as active when the current location is `/user/roles`. The way I've described it in the guide would seem to imply that only ancestors can be active.

When I was writing my description of how 'active' is defined, I stayed clear of mentioning `route.matched`, because I didn't think most readers would know what that was, but in my mind they are closely related concepts. The edge case above is one instance where a link is shown as 'active' despite the route not being in `route.matched`.

skirtles-code:
> I'm wondering if we should name the page something else that includes active/matching. We could even omit the RouterLink keyword. So maybe, just "Active Links" works better. What do you think?

Yeah, I think that makes sense. I've also renamed the first section heading so it doesn't clash with the main heading.

posva:
Ah i see now what you meant. It would be inconvenient if it was the other way around but it doesn’t for against the idea of the matched array

posva (approved):

Issue #2187, opened 2024-03-25
hash decode not match the later encode


zcqno1:
### Reproduction

https://jsfiddle.net/qhn7gka8/

### Steps to reproduce the bug

1. route like `/test#/?redirect=%2F%3Fid%3D1%23%2Fabc` (just `` `/test#/?redirect=${encodeURIComponent('/?id=1#/abc')}` ``)
2. do replace the upper route, the final route.fullpath turn to `/test#/?redirect=/?id=1#/abc`, that's unexpected
3. `push` works well.

### Expected behavior

the hash of route remain its original encode

### Actual behavior

hash's original encode lost

### Additional information

posible reason.

The hash part of route is decode in the `parseURL` function. 

https://github.com/vuejs/router/blob/4e82939d3094569fa732937834544ec76aa7a014/packages/router/src/location.ts#L89

`decode` uses `decodeURIComponent`, which would decode the `/?`.

After doing a `replace` action, it calls `parseURL` first, and then calls `resolve(to)` later.

https://github.com/vuejs/router/blob/4e82939d3094569fa732937834544ec76aa7a014/packages/router/src/router.ts#L657

In  function `resolve`, hash is encode. 
However, `encodeHash` use `encodeURI` indexof `encodeURIComponent`, so the hash would never be encoded to the original hash if there's characters `/?` in it.

https://github.com/vuejs/router/blob/4e82939d3094569fa732937834544ec76aa7a014/packages/router/src/router.ts#L531

https://github.com/vuejs/router/blob/4e82939d3094569fa732937834544ec76aa7a014/packages/router/src/encoding.ts#L72

```js
/**
 * Encode characters that need to be encoded on the path, search and hash
 * sections of the URL.
 *
 * @internal
 * @param text - string to encode
 * @returns encoded string
 */
function commonEncode(text: string | number): string {
  return encodeURI('' + text)
    .replace(ENC_PIPE_RE, '|')
    .replace(ENC_BRACKET_OPEN_RE, '[')
    .replace(ENC_BRACKET_CLOSE_RE, ']')
}

/**
 * Encode characters that need to be encoded on the hash section of the URL.
 *
 * @param text - string to encode
 * @returns encoded string
 */
export function encodeHash(text: string): string {
  return commonEncode(text)
    .replace(ENC_CURLY_OPEN_RE, '{')
    .replace(ENC_CURLY_CLOSE_RE, '}')
    .replace(ENC_CARET_RE, '^')
}
```

[...]

PR #2189, opened 2024-03-25
fix: avoid normalizing the fullPath

posva:
Close vuejs/router#2187

### DIFF
--- a/packages/router/__tests__/router.spec.ts
+++ b/packages/router/__tests__/router.spec.ts
@@ -534,6 +534,18 @@ describe('Router', () => {
     })
   })
 
+  // https://github.com/vuejs/router/issues/2187
+  it('keeps a consistent value on fullPath when resolving', async () => {
+    const { router } = await newRouter()
+    const targetLoc = '/search#/?redirect=%2F%3Fid%3D1%23%2Fabc'
+    expect(router.resolve(targetLoc).fullPath).toBe(targetLoc)
+    await router.push(targetLoc)
+    expect(router.currentRoute.value.fullPath).toBe(targetLoc)
+    await router.push('/')
+    await router.replace(targetLoc)
+    expect(router.currentRoute.value.fullPath).toBe(targetLoc)
+  })
+
   describe('navigation cancelled', () => {
     async function checkNavigationCancelledOnPush(
       target?: RouteLocationRaw | false
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -525,13 +525,19 @@ export function createRouter(options: RouterOptions): Router {
     // we need to run the decoding again
     matchedRoute.params = normalizeParams(decodeParams(matchedRoute.params))
 
-    const fullPath = stringifyURL(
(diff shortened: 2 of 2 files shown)


dargmuesli:
Hmm, I suspect this change to break some routing in an i18n'd nuxt app of mine: https://github.com/dargmuesli/creal/pull/451. I get no error though, so I'm not sure if another library needs to adapt. Just want to point this out as meta info, let's see if someone creates a more specific issue faster than me.

Explanation on the failling pipeline: a playwright test fails which checks for a router navigation to happen when a different language is selected in the website's footer. So if you want to test it yourself, you can checkout that branch and try to change the language in the footer. WIth  `vue-router` `v4.3.1` nothing happens, `v4.3.0` works.

posva:
Thanks for the info, if you create a boiled down repro without external deps that would be really helpful

posva:
Yes, a reproduction without i18n and Nuxt would help. The repro with i18n still looks like a black box without any errors so I can't know what is happening

BobbieGoede:
I understand this isn't particularly helpful for debugging but I figured this would at least demonstrate this release contains breaking changes.

Hopefully I have time to dig into why this is breaking in our case and provide a more minimal reproduction, in the meantime Nuxt i18n users will have to pin vue-router to `4.3.0`.

DavyJones21:
Hi.

Here's an example:
From the api server comes the path, like this: **/custom/path**.

I throw path localization in the custom router-link component (I use my own logic without vue-i18n):
`<AppRouterLink :to="/custom/path" />`

Inside AppRouterLink component:
`const localizedTo = computed(() => {
    return {
        ...to,
        params: {
            ...to?.params,
            locale: locale.value,
        },
    }
})`

`<RouterLink :to="localizedTo">...</RouterLink>`

Instead of **/en/custom/path** I get **/custom/path** and when I go to such a link the router does not work and the site crashes. With version 4.3.0 everything works fine.

BobbieGoede:
I figured out what was causing this to break in the i18n module and have a fix lined up (https://github.com/nuxt-modules/i18n/pull/2922) that should resolve the issue for us and will work with both vue-router `4.3.0` and `4.3.1`.

In our case we were resolving localized routes by replacing a resolved route's `name`, since the resolved route still had the previous `fullPath` that would be prioritized instead as per this PR.

@DavyJones21 
I'm curious to see your own logic for resolving localized routes, since you're experiencing the same issue I'm guessing your localized route resolution is either the same or very similar, you may be able to fix this issue in the same way (https://github.com/nuxt-modules/i18n/pull/2922/files#diff-35b10ac721121ca10ffabe8186f63f1e7abfef49564ec90c7ab23bf24d3d3870)

posva:
I reverted this and released a new patch

DavyJones21:
@BobbieGoede 
Yes, removing the fullPath from the resolved router object does solve the problem. It is strange that subsequent **params** changes do not work as they should.

Alternatively, I can add the locale to the path so that the resolving is immediately successful.

Thanks for help.

Issue #2190, opened 2024-03-28
Inconsistent trailing slash behaviour with  multiple optional parameters


panstromek:
### Reproduction

https://jsfiddle.net/7sg5pry1/

### Steps to reproduce the bug

Try route like this:

`'/page/:first(a)?:second(b)?'`

(clicking the button in the reporduction link)

### Expected behavior

It matches:

`/page`

Text "should be visible" is visible in  repro

### Actual behavior

Doesn't match `/page`

Text "should be visible"  is not visible in  repro

### Additional information

This is inconsistent with the behaviour with only a single optional parameter:

`'/page/:first(a)?'` matches `/page`

[Documentation](https://router.vuejs.org/guide/essentials/route-matching-syntax.html#Sensitive-and-strict-route-options) doesn't explicitly mention this case, but implies should that trailing slash is removed in both cases

posva:
By having 2 parameters, the parser expects the path to have at least one character. It only makes it fully optional if it's **only** an optional param. So you can group both params into a single one or create an alias without the params. I'm curious to see what the real route is

panstromek:
The real route is something like this: `'/articles/:slug([\\w-]+-)?:article_id(\\d+)'`. The idea is to have optional slug for SEO and append id to it, so it should match

`/articles/99`
but also
`/articles/new-vue-version-released-99`

panstromek:
I realized this example doesn't have optional id, sorry. That one is something like:

`'users/:userId/videos/:slug([\\w-]+-)?:video_id(\\d+)?'`

which shows a list of videos in swipe screen, and optionally starts at the specific one based on Id

posva:
I see. Thanks for sharing! Using an alias is indeed the correct solution

panstromek:
Ok, it would be useful to mention this in the docs, though. I broke a production for a few minutes, it's pretty unexpected behaviour. I'll submit a PR when I have some time to look into it

PR #2192, opened 2024-03-29
docs: mention an edge case with mutliple optional params in the docs

panstromek:
See #2190 for more info.

### DIFF
--- a/packages/docs/guide/essentials/route-matching-syntax.md
+++ b/packages/docs/guide/essentials/route-matching-syntax.md
@@ -114,6 +114,15 @@ const routes = [
 
 Note that `*` technically also marks a parameter as optional but `?` parameters cannot be repeated.
 
+When using multiple optional parameters in succession, the behavior is slightly different because the parser expects at least one of them to match.
+
+```js
+const routes = [
+  // This will match /users/42posva, /users/42, /users/posva but not /users,  
+  { path: '/users/:userId(\\d+)?:name(\\w+)?' },
+]
+```
+
 ## Debugging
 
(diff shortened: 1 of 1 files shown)


packages/docs/guide/essentials/route-matching-syntax.md
```diff
@@ -114,6 +114,15 @@ const routes = [
...
 Note that `*` technically also marks a parameter as optional but `?` parameters cannot be repeated.
 
+When using multiple optional parameters in succession, the behavior is slightly different because the parser expects at least one of them to match.
+
+```js
+const routes = [
+  // This will match /users/42posva, /users/42, /users/posva but not /users,  
+  { path: '/users/:userId(\\d+)?:name(\\w+)?' },
+]
+```
```
posva, line 124:
```suggestion
If the route segment contains more than **just an optional parameter**, it won't match a path **without the trailing slash**. For example:

- `/users/:uid?-:name?` won't match `/users`, only `/users/-` or even `/users/-/`
- `/users/:uid(\\d+)?:name? won't match `/users`, only `/users/`, `/users/2`, `/users/2/`, etc

You can play around with the matching syntax [in the playground](https://paths.esm.dev/?p=AAMsIPQg4AoKzidgQFoEXAmw-IEBBRYYOE0SkABTASiz1qgBpgQA1QTsFjAb3h2onsmlAmGIFsCXjXh4AIA.&t=/users/2/#)
```

posva (approved):
Thanks! I changed it a bit to be more general

PR #2235, opened 2024-05-12
docs: adds Bun package manager option in Installation Guide

AdrianDamiao:
While I was reading the documentation for creating a Vue 3 project using Vue Router, I realized that this option was missing and it was possible to default everything to the latest version, just like NPM. This PR adds the Bun package manager as an option as well as the Vuejs.org documentation

image


posva:
Duplicate of https://github.com/vuejs/router/pull/2089

PR #1847, opened 2023-05-13
build: add vue-router.esm-browser.prod.js minified file

kianfang:
Build vue-router.esm-browser.prod.js midified file.
We can access this CDN resource in browser after publish:
[https://unpkg.com/vue-router/dist/vue-router.esm-browser.prod.js](https://unpkg.com/vue-router/dist/vue-router.esm-browser.prod.js)

image

PR #2242, opened 2024-05-17
build: add vue-router.esm-browser.prod.js

skirtles-code:
Based on #1847.

Currently we have `vue-router.esm-browser.js`, but there isn't a production version of that file.

The key benefits of the newly added `vue-router.esm-browser.prod.js` are:

1. It's minified.
2. It doesn't have a dependency on `@vue/devtools-api`.

That second point is important, as various CDNs can automatically apply the minification, but they can't strip out the dependency on `@vue/devtools-api`.

The `moduleSideEffects: false` setting is responsible for ensuring `@vue/devtools-api` is removed. Without that, we end up with `import "@vue/devtools-api"` in the built file. See https://rollupjs.org/configuration-options/#treeshake-modulesideeffects for details of that setting. While the setting is being applied to all builds, none of the other builds are actually impacted in practice.

Do we actually *need* a production build for `esm-browser`? Not sure, but based on the comments on #1847 and #2241, it does seem that some people would find it useful. An equivalent file is already provided by both Vue Router 3 and Vue core.

Closes #1847.


posva (approved):
Thanks a lot for this!

Issue #2266, opened 2024-06-11
 [Vue Router warn]: Finding ancestor route "/:path(.*)*" failed for "/:path(.*)*"


rex-ll:
### Reproduction

RT

### Steps to reproduce the bug

After upgrading to the latest version

`
{
  path: '/:path(.*)*',
  name: PAGE_NOT_FOUND_NAME,
  component: LAYOUT,
  meta: {
    title: 'ErrorPage',
  },
  children: [
    {
      path: '/:path(.*)*',
      name: PAGE_NOT_FOUND_NAME,
      component: EXCEPTION_COMPONENT,
      meta: {
        title: 'ErrorPage',
      },
    },
  ],
}
`

### Expected behavior

no warning

### Actual behavior

warning log

skirtles-code:
I think I managed to reproduce this:

[...]

PR #2267, opened 2024-06-11
feat: throw if parent and child routes have the same name

skirtles-code:
#2266 demonstrated an edge case that can lead to inconsistent internal state in the matcher.

The problem arises when using nested routes. If the child (or deeper descendant) has the same `name` as its parent, the parent is removed at the same time that the child is added.

Removing a parent route should also remove its children, but in this case that removal happens just before the child is added to the matcher. The child still gets added to the matcher, and its `parent` property still points to the parent route, even though that has now been removed. The parent can no longer be resolved directly, but when resolving the child, the `matched` array will still contain the removed parent.

This problem can occur during the initial router creation:

```js
createRouter({
  // ...
  routes: [{
      name: 'abc',
      // ...
      children: [{
        name: 'abc',
        // ...      
      }]
  }]
})
```

It can also happen when adding a route dynamically:

```js
const router = createRouter({
  // ...
  routes: [{
      name: 'abc',
      // ...
  }]
})

router.addRoute('abc', {
  name: 'abc',
  // ...      
})
```

This PR introduces a dev-only check for ancestors with the same name and throws an error if a duplicate is found.

### DIFF
--- a/packages/router/src/matcher/index.ts
+++ b/packages/router/src/matcher/index.ts
@@ -512,7 +512,7 @@ function checkSameNameAsAncestor(
       throw new Error(
         `A route named "${String(record.name)}" has been added as a ${
           parent === ancestor ? 'child' : 'descendant'
-        } of a route with the same name. This is not allowed, a nested route must not use the same name as an ancestor.`
+        } of a route with the same name. Route names must be unique and a nested route cannot use the same name as an ancestor.`
       )
     }
   }
--- a/packages/router/src/matcher/index.ts
+++ b/packages/router/src/matcher/index.ts
@@ -154,8 +154,12 @@ export function createRouterMatcher(
 
         // remove the route if named and only for the top record (avoid in nested calls)
         // this works because the original record is the first one
-        if (isRootAdd && record.name && !isAliasRecord(matcher))
+        if (isRootAdd && record.name && !isAliasRecord(matcher)) {
+          if (__DEV__) {
+            checkSameNameAsAncestor(record, parent)
+          }
           removeRoute(record.name)
+        }
       }
 
       // Avoid adding a record that doesn't display anything. This allows passing through records without a component to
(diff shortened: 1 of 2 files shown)


sherleysong:
How can i make / to Home , make /foo to Foo , while all the left "/:pathMatch(.*)" to PAGE_NOT_FOUND ?
` {  path: '/',
        name: 'home',
        component: Main,
        redirect: '/home',
        children: [
          {
            path: '/home',
            name: 'home',
            component: Home
          },
          {
            path: '/:pathMatch(.*)',
            component: () => Error
          }
        ] }`

skirtles-code:
@sherleysong I don't fully understand your question, but you need to remove the `name: 'home'` from the parent route.

sherleysong:
> @sherleysong I don't fully understand your question, but you need to remove the `name: 'home'` from the parent route.

It worked , thank you !   

The running code was    {  path: '/', name: 'home', component: Main, redirect: '/home', children: [ { path: '/home', name: 'home', component: Home }, { path: '/:pathMatch(.*)', component: () => Error } ] } 

@vue-router@4.3.2,  when we visit  www.xxx.cn  ,  it will redirect  to Home. 
@vue-router@4.3.3,  when we visit  www.xxx.cn  ,  it will redirect  to  Error.   (which was wrong yesterday)

as you recommended , i removed the `name: 'home'` from the parent route ,  then 
@vue-router@4.3.3,  when we visit  www.xxx.cn  ,  it will redirect  to  Home.  

Anyway , it worked for me , thank you again!

packages/router/src/matcher/index.ts
```diff
@@ -499,6 +503,21 @@ function checkChildMissingNameWithEmptyPath(
...
+function checkSameNameAsAncestor(
+  record: RouteRecordRaw,
+  parent?: RouteRecordMatcher
+) {
+  for (let ancestor = parent; ancestor; ancestor = ancestor.parent) {
+    if (ancestor.record.name === record.name) {
+      throw new Error(
+        `A route named "${String(record.name)}" has been added as a ${
+          parent === ancestor ? 'child' : 'descendant'
+        } of a route with the same name. This is not allowed, a nested route must not use the same name as an ancestor.`
```
posva, line 515:
```suggestion
        } of a route with the same name. Route names must be unique and a nested route cannot use the same name as an ancestor.`
```

posva (approved):

posva:
Thanks for this. I really didn't expect people to fall into this but I was wrong

PR #2291, opened 2024-07-04
Fix: do not encode slashes in catch-all routes

### LINKED ISSUE #1638: Using router.push or .replace will cause forward slashes in url to be wrongly escaped when using regex to match path
### Version
4.1.6

### Reproduction link
[codesandbox.io](https://codesandbox.io/s/hello-vite-forked-vjoci2?file=/src/main.js)

### Steps to reproduce
1. Load the codesandbox.io project using the link provided.
2. Click on About page.
3. Click on the "Add hash" button, this will call `router.replace({ hash: '#something' })`.
4. Notice that vue-router correctly appends the # in the url. `(https://domain.com/about.html#yippy)`.
5. Click on Deep About page.
6. Click on the "Add hash" button.
7. Notice that the url becomes `https://domain.com/test/deep%2Fabout.html#yaya`.

### What is expected?
Url to become `https://domain.com/test/deep/about.html#yaya`

Url became `https://domain.com/test/deep%2Fabout.html#yaya`

image

---
This is for the front-end of a CMS-supported project, where the prefix of the URL is controlled by the CMS and it varies depending on the environment. Hence the need to first match it using `:pathMatch()` then match the actual page to render using `:pageName()`.

Also, all pages has to be suffixed with .html, hence I am using a (kind of) complex regex to match the page. It works, however,  I stumble upon this issue.

ll3006:
This fixes #1638

Encoding of the parameters is now done by the matcher and is dependend on the matcher keys: every key has a keepSlash property that reflects if it should encode the slash character or not.

The keepSlash currently applies to all custom regex that either match a slash or contain one in their literal part, excluding lookaheads, lookbehinds and negative ranges. This is very simplistic but should be enough to cover both catch-all (e.g. "(.\*)" routes) and explicit path regexes (e.g. "(deep/path/[a-z]\*/)" )

I have also added a few tests to ensure everyting is working properly.
 
The Router > 'can redirect to a star route when encoding the param' test is currently failing as it sets a parameter with an urlencoded slash (%2F) and expects to find it as it is in the route. However, since this PR allows literal slashes in star routes, the actual url contains the sequence as a literal (%252F) rather than keeping the escaped slash. 

I could edit the test, but I am unsure if this is behaivour that is being relied upon. In case it is, more work is needed to distinguish user-set urlencoded path segments from the normal usecase.

To test this PR I have updated the sandbox from #1638 here: [codesandbox.io](https://codesandbox.io/p/devbox/vue-router-catch-all-slash-gjxc9w)


posva:
Thanks, this will be added once route resolvers become the default:

- https://github.com/vuejs/router/pull/2415

PR #2295, opened 2024-07-07
fix: router typescript definations related to vue package

lovetingyuan:
image
image

after

image
image

I noticed the comment : `// TODO: figure out why it cannot be 'vue' like said in docs`, actullay, it works in the latest version of vue(3.4.31)


posva (approved):
Thanks. This can be changed now indeed. It was already updated in pinia

bjacobgordon-cy:
@posva This breaks type parameters for in-component navigation guards supplied to `defineComponent`.

PR #2296, opened 2024-07-08
docs: reorder symbols

cwj0417:
reorder 2 markdown symbols that I think are more appropriate.

### DIFF
--- a/packages/docs/guide/essentials/route-matching-syntax.md
+++ b/packages/docs/guide/essentials/route-matching-syntax.md
@@ -24,7 +24,7 @@ const routes = [
 ]
 ```
 
-But in some scenarios we don't want to add that static section `/o``/p`. However, `orderId` is always a number while `productName` can be anything, so we can specify a custom regex for a param in parentheses:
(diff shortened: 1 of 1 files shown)


packages/docs/guide/essentials/route-matching-syntax.md
```diff
@@ -24,7 +24,7 @@ const routes = [
 ]
 ```
 
-But in some scenarios we don't want to add that static section `/o`/`p`. However, `orderId` is always a number while `productName` can be anything, so we can specify a custom regex for a param in parentheses:
+But in some scenarios we don't want to add that static section `/o``/p`. However, `orderId` is always a number while `productName` can be anything, so we can specify a custom regex for a param in parentheses:
```
posva, line 27:
```suggestion
But in some scenarios, we don't want to add that static section `/o` or `/p`. However, `orderId` is always a number while `productName` can be anything so we can specify a custom regex for a param in parentheses:
```

posva (approved):
Thanks!

PR #2305, opened 2024-07-17
docs: fix example in passing-props.md

rngesus-wept:
I think the example given for passing in `id` through props is incorrect. I needed to assign the return value of `defineProps` and then reference that in order to get the right behavior.

Vue 3.4.29, Vue-Router 4.3.3

### DIFF
--- a/packages/docs/guide/essentials/passing-props.md
+++ b/packages/docs/guide/essentials/passing-props.md
@@ -36,14 +36,14 @@ We can remove the direct dependency on `$route` in `User.vue` by declaring a pro
 ```vue [Composition API]
 <!-- User.vue -->
 <script setup>
-defineProps({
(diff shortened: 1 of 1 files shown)


skirtles-code:
The example appears to be correct to me. It shouldn't be necessary to use `props.id` in the template. This isn't specifically a Vue Router thing, accessing props like this is part of Vue core.

Perhaps you already have something called `id` in the same scope that's clashing with the prop?

posva:
Thanks for the pr but it’s what @skirtles-code said

rngesus-wept:
Ah, okay, thanks for the hint. I double-checked and using `{{ id }}` without other qualification in the template works fine, but referring to it anywhere in *code* requires a reference to `props`. Sorry for the trouble.

PR #2313, opened 2024-07-26
docs: add bun command examples

dannote:
It's time!


skirtles-code:
Similar to #2089.

I understand why that PR was closed, but the core Vue docs do include similar instructions for `bun` at <https://vuejs.org/guide/quick-start.html>.

PR #2326, opened 2024-08-04
chore: use `with` instead of `assert` in rollup.mjs

negezor:
Fixes a build error on newer versions of Node.js. Import attributes by specification should use `with` keyword instead of `assert` and it's already on stage 3, so proposal is accepted. https://github.com/tc39/proposal-import-attributes

This is also mentioned in TypeScript 5.3 https://devblogs.microsoft.com/typescript/announcing-typescript-5-3/#import-attributes

```
Scope: 3 of 4 workspace projects
packages/router build$ rimraf dist && rollup -c rollup.config.mjs
│ [!] SyntaxError: Unexpected identifier 'assert'
```


posva (approved):
Thanks!

PR #2333, opened 2024-08-06
revert ComponentCustomOptions and ComponentCustomProperties extensions (fix #2332)

bjacobgordon-cy:
Partially reverts https://github.com/vuejs/router/pull/2295
- [Demonstration of changes](https://vimeo.com/995552066/753e61ef9e?share=copy)

### DIFF
--- a/packages/router/src/globalExtensions.ts
+++ b/packages/router/src/globalExtensions.ts
@@ -8,11 +8,10 @@ import type { RouterLink } from './RouterLink'
 import type { Router } from './router'
 import type { TypesConfig } from './config'
 
-/**
- * NOTE: this used to be `@vue/runtime-core` but it should have been `vue` for a long time. Using both declaration at
- * the same time breaks so using only one everywhere is the preferred way.
- */
-declare module 'vue' {
+declare module '@vue/runtime-core' {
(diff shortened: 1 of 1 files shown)


Brads3290:
I noticed the same across several of my Nuxt apps after updating dependencies. Pinning to 4.4.0 fixed the issue.

skywalker512:
In my experience, if the dependencies in the project are written as `declare module 'vue'` or `declare module '@vue/runtime-core'` (Both ways of writing appear in your dependence.),

ComponentCustomProperties will be completely unavailable.(I'm not familiar with Typescript) 

You can check whether your dependency has such a situation. When I completely replace it with `declare module 'vue'`, the problem disappears. Then we can see that pinia also makes a similar replacement.

https://github.com/vuejs/pinia/commit/8a6ce86db83b6315c067c8a98c898b3c74efe62e

danielroe (approved):
Created a reproduction of the issue at https://github.com/vuejs/router/issues/2341.

bjacobgordon-cy:
@danielroe @posva I believe I've figured out what was missing when augmenting the `vue` module. What do you think?

aiqin-bao:
So what's the ultimate solution?

PR #2377, opened 2024-09-30
fix: custom directive doesn't work when custom=true

### LINKED ISSUE #2375: Directives do not work on <RouterLink custom>
### Reproduction

https://play.vuejs.org/#eNqVVNtu2zAM/RXBG+AUaKxdij4YSZB1K9YNu6Ebtod5D47NJGpkyZDoJEWQfx8lJY7TdRn2kAvJw8NDitImqnKhkjsbpZGoam2QbVhhIEd4Vddsy6ZGVyxeNhBnqgUY3SCYNprw4DhAXO4+ltd1EvIz1TL36HOWKcaSxkIvpAe70o3CXvyE0uKz6DzapZO+gS2MqJFZwKYetbU+CLV4q3XZFtw7dlU7sKv8GEX2DlRoZZEtr7Qs2ZCBZMMRfScW7yUkU63wB4jZHCkWTwhDGQMe5JAQMhCqWlJrZDE2aCWhHmYRzyK27Ls05vnIVWipTcoMlFk0eiMMFCiWwFbaLOygbSCQTcyB1DXwn5ylBqti9NyBmkiIccBbzTTlcAIJPrIHH6HS5v5GWKSf853z9ngDaIj9dgfakd/oCr4LWLUz3zvafYC1B4bp77ZqeFSit3HNz0P19DFJvbNzB/HZNmU/ncFIfZ3jPGUxj0mzJkEKFKYHTVuf9Yu+t7R4NIKuuIfbxmSuZjRjtDTb02c/fzHaMw04GQ8nHWbTr/Kabp1WVMl3mO0CVCBl3uN8h7k6dxbNEWubct6oejFLqC1+QIwvkpfJM17SVDreBGzVnxi9svT/jth92457TCBewhK1lraf1+JvJf4Aji+Ty+Q5l2LCiZ0LVcLaczvqLQ2U2kRLhzoVswdNupMQEsznGgUd+lGzuZR69d770DTQCi3mUCwe8d/ZdZD8xQB1t4ROc5ibGWAIX3/9BGv63wYrXTaS0CeCt2C1bJzGALtqVEmyOziv9p0/M6Fm3+z1GkHZfVNOqJ+Gx/uDfH2i9YPcl8lFZ4rdp8zt5PGqhRviMN4mj5UaeQjyo2hnB/2yd16/U7ysaOiOVfTUOGa6ABs2NzA9ZypfihklkFB3IXzxnKUuSCj3Qw/UuJCiWJC9R4dHi/aFfE9zRBOuk8s+Eu/k5/9uY/sbfE5I2w==

Click the reproduction link.

### Expected behavior

The second link should be bold.

### Actual behavior

The second link is not bold.

### Additional information

I understand why this happens. Vue 3 does not support custom directives on components that render multiple root nodes. This is also the case if the render function returns an array of a single vnode since it'll get converted into a fragment.

Screenshot 2024-09-29 at 11 23 22 am

This is the RouterLink render function:

https://github.com/vuejs/router/blob/14219b01bee142423265a3aaacd1eac0dcc95071/packages/router/src/RouterLink.ts#L309-L327

If `custom` is false then a single `<a>` element is the root element of the component and everything works as expected.

If `custom` is true then the result of calling the default slot function is returned instead. The vue template compiler will wrap the default slot in an array even if it only contains a single node, so it will become a fragment.

The only fix I can think of is to change the RouterLink render function so that the result of the default slot is unwrapped if it is an array of length 1:

```js
function preferSingleVNode(vnodes) {
  return Array.isArray(vnodes) && vnodes.length === 1 ? vnodes[0] : vnodes;
}
```

```js
const children = preferSingleVNode(slots.default && slots.default(link))
```

It's really unfortunate that Vue 3 cannot do something that Vue 2 was able to do: https://github.com/vuejs/core/issues/3914

decademoon:
Fix #2375.

### DIFF
--- a/packages/router/__tests__/RouterLink.spec.ts
+++ b/packages/router/__tests__/RouterLink.spec.ts
@@ -911,7 +911,7 @@ describe('RouterLink', () => {
         `,
         components: { RouterLink },
         directives: { Directive },
-        name: 'AppLink'
+        name: 'AppLink',
       })
 
       const { wrapper } = await factory(
--- a/packages/router/__tests__/RouterLink.spec.ts
+++ b/packages/router/__tests__/RouterLink.spec.ts
@@ -358,7 +358,8 @@ async function factory(
   currentLocation: RouteLocationNormalized,
   propsData: any,
   resolvedLocation: RouteLocationResolved,
-  slotTemplate: string = ''
+  slotTemplate: string = '',
+  component: any = RouterLink
 ) {
   const route = createMockedRoute(currentLocation)
   const router = {
@@ -373,7 +374,7 @@ async function factory(
   }
   router.resolve.mockReturnValueOnce(resolvedLocation)
 
-  const wrapper = mount(RouterLink as any, {
+  const wrapper = mount(component, {
     propsData,
     global: {
       provide: {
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #2410, opened 2024-11-25
docs: fix hash anchor link for `RouteLocationRaw`

DDDDD12138:
### Descriprion

This PR addresses an issue in the Chinese documentation where the links related to `RouteLocationRaw` do not correctly navigate to the intended sections. The problem arises due to incorrect hash anchors that do not align with the corresponding headings.

|Incorrect|Correct|
|---|---|
|image| image|


posva:
I just updated the API generation. Could you give this a new look please?

posva:
The API is generated and I think the copy in Chinese was never updated to the latest copy

DDDDD12138:
Thank you for your response. 
Is it possible to implement an approach similar to i18n in the future? If there is no translation available, it can fallback to displaying in English. This way, even in other language environments without translations, the latest documentation can still be used.

posva:
Unfortunately not. It's up to the community to come up with the API translation and to keep it up to date

PR #2413, opened 2024-12-05
docs: incorrect description of content in the dynamic-routing section

### LINKED ISSUE #2412: The description of the "Adding routes inside navigation guards" section in the document is incorrect
### Reproduction

Docs incorrect

1.The description of the "Adding routes inside navigation guards" section in the document is incorrect
2.The incorrect content is as follows
"Second, hasNecessaryRoute() returns false after adding the new route to avoid an infinite redirection."

### Expected behavior

The expected correct information is "Second, hasNecessaryRoute() returns **true** after adding the new route to avoid an infinite redirection."
hasNecessaryRoute() returns true.

### Actual behavior

The incorrect content is as follows
"Second, hasNecessaryRoute() returns false after adding the new route to avoid an infinite redirection."

### Additional information

_No response_

stonelee2048:
Close #2412


posva (approved):
Thanks!
Open directly the PR next time, without an issue, it's faster 😄

PR #2433, opened 2025-01-03
remove unused param.

zhangenming:
Or some other purpose I don't get?


posva:
Thanks but this is used for types (https://uvr.esm.is)

PR #2434, opened 2025-01-07
fix(resolve): exclude empty optional arguments from params

### LINKED ISSUE #2419: Active class not applied on `RouterLink` when using `to` with `string` or object with `path` key
### Reproduction

https://stackblitz.com/edit/vitejs-vite-a1m7gbnq

1. Open [the reproduction link](https://stackblitz.com/edit/vitejs-vite-a1m7gbnq)
2. Click on third item (`Go to Feature one`)
3. Check the active classes applied to the first and second links.

### Expected behavior

The `RouterLink` component with the `/features` value is expected to have the active class (`router-link-active`) applied when the current route is `/features/one`, regardless of how the `to` prop value is provided.

### Actual behavior

The active class (`router-link-active`) is applied only when the `to` prop is passed with a route `name`. However, for another `RouterLink` where the `to` prop is provided as a string, the active class is not applied.

### Additional information

There is another difference between these two `RouterLink` components, though I'm not sure if it's a bug:
```
<RouterLink :to="{ name: 'features' }">
<RouterLink to="/features">
```

Clicking only on the **second** link causes the `$route.params` to include a `pathMatch` key.
You can verify this by clicking on the first two links in [the reproduction link](https://stackblitz.com/edit/vitejs-vite-a1m7gbnq).

babu-ch:
fix https://github.com/vuejs/router/issues/2419

Removed if optional parameter is empty
Now the param will match if :to is string and if it is object

In the case of string path, it should not be an empty string parameter(Right?)
, so I thought there would be no problem with excluding it.

However, this may be a breaking change if you have users who were using this

In addition, I added a router-link to Playgound to check the operation.

## Summary by CodeRabbit

* **New Features**
  * Added navigation links and a route to access feature pages, including nested feature paths.

* **Bug Fixes**
  * Fixed resolution so missing optional route parameters are omitted instead of returning empty placeholders.

* **Tests**
  * Added tests covering optional-parameter resolution, including catch-all nested paths and differences when resolving by path vs name.

### DIFF
--- a/packages/router/__tests__/matcher/resolve.spec.ts
+++ b/packages/router/__tests__/matcher/resolve.spec.ts
@@ -685,12 +685,12 @@ describe('RouterMatcher.resolve', () => {
       assertRecordMatch(
         { path: '/:a?', components, name: 'a' },
         { path: '/' },
-        { path: '/', params: { a: '' }, name: 'a' }
+        { path: '/', params: {}, name: 'a' }
       )
       assertRecordMatch(
         { path: '/a/:a?', components, name: 'a' },
         { path: '/a/' },
-        { path: '/a/', params: { a: '' }, name: 'a' }
+        { path: '/a/', params: {}, name: 'a' }
       )
     })
 
--- a/packages/router/__tests__/matcher/resolve.spec.ts
+++ b/packages/router/__tests__/matcher/resolve.spec.ts
@@ -888,6 +888,29 @@ describe('RouterMatcher.resolve', () => {
         { name: 'h', path: '/', params: {} }
       )
     })
+
+    // #2419 (skirtles-code's comment): `{ params: { id: undefined } }`
+    // should resolve the same as omitting the key entirely so active-link
+    // comparisons stay consistent. `null` is a type error and is not
+    // supported (see commented-out case below).
+    it('treats explicit undefined optional params as absent', () => {
+      const record = {
+        path: '/features/:id?',
+        name: 'features',
+        components,
+      }
+      assertRecordMatch(
+        record,
+        { name: 'features', params: { id: undefined } },
+        { name: 'features', path: '/features', params: {} }
+      )
+      // null is rejected by the types and not supported at runtime:
+      // assertRecordMatch(
+      //   record,
+      //   { name: 'features', params: { id: null } },
+      //   { name: 'features', path: '/features', params: {} }
+      // )
+    })
   })
 
   describe('LocationAsRelative', () => {
--- a/packages/router/__tests__/matcher/resolve.spec.ts
+++ b/packages/router/__tests__/matcher/resolve.spec.ts
@@ -694,6 +694,25 @@ describe('RouterMatcher.resolve', () => {
       )
     })
 
+    // #2419: resolving by string path must match the same-named route's
+    // missing-optional-params behavior so active-link comparisons stay in sync
+    it('drops missing optional splat when resolving by path', () => {
+      assertRecordMatch(
+        { path: '/features/:pathMatch(.*)*', components, name: 'features' },
+        { path: '/features' },
+        { path: '/features', params: {}, name: 'features' }
+      )
+      assertRecordMatch(
+        { path: '/features/:pathMatch(.*)*', components, name: 'features' },
+        { path: '/features/one' },
+        {
+          path: '/features/one',
+          params: { pathMatch: ['one'] },
+          name: 'features',
+        }
+      )
+    })
+
     it('keeps required trailing slash (strict: true)', () => {
       const record = {
         path: '/home/',
--- a/packages/router/__tests__/matcher/resolve.spec.ts
+++ b/packages/router/__tests__/matcher/resolve.spec.ts
@@ -888,29 +888,6 @@ describe('RouterMatcher.resolve', () => {
         { name: 'h', path: '/', params: {} }
       )
     })
-
-    // #2419 (skirtles-code's comment): `{ params: { id: undefined } }`
-    // should resolve the same as omitting the key entirely so active-link
-    // comparisons stay consistent. `null` is a type error and is not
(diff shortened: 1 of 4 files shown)


packages/router/src/matcher/index.ts
```diff
@@ -310,6 +310,12 @@ export function createRouterMatcher(
         // we know the matcher works because we tested the regexp
         params = matcher.parse(path)!
         name = matcher.record.name
+
+        matcher?.keys.forEach(key => {
```
skirtles-code, line 314: Can the `?.` be changed to `.` instead?

babu-ch: @skirtles-code 
Thank you for your comment.
Oh, it was a remnant from writing outside of this block.
Fixed!
https://github.com/vuejs/router/pull/2434/commits/66f824f7f9a2b8f71828267b32ec6a14ce60d7e4

skirtles-code:
I was wondering about some other edge cases. I'm not sure whether these necessarily impact this PR, but I thought they were worth considered before it's merged.

I put together this Playground to demonstrate:

- [Playground](https://play.vuejs.org/#eNqtVdtu2zgQ/RWuuoAcwBYdp9sHrepNt+hudrGXoi36UvWBlmibtUQSJOU4MPzvHV50seOmQREEScSZMzfOzOE+qgnjyRcdpRGrpVAG7VGhKDH0lZTogJZK1CjeNjTOeQAo0RiqgibB/tirrV2rI1Im3jbnndcR/F7kHKGk0XTkzf25Fg03o/gZmMUX0TgK5pBbZmgtK7CfW2C2vpzf0KoSNthPGYajE0v3Dz60UYKv5q8bpSgPGSNJzDrNcNCh/R797BTJsqmqt6BEh4Nzg72fjJNt6/Cdy/IfxjfIiJd5hPNo/qeAb4Qz3Cu/CV9C7Y2iujdrJd8xT539HnFS0xTFrVWMDp2rc8of9DqGS1Kk1ik4ZWWKGl7SJeO0hEl4KOD4FP1E8Tl05lGhPfCJosZwgY+IaWFPFfFydvXIoB55L26G22nN7EYf5/KR0VuEAy6oMzxYKThqc1fZz8Qv5KQCzxNSGLalaG8tC1EJlSJFy19zDotyAqU7AB8ZLEixWQGEQ9p3dltvvaFdQRsLFtw7OEs//9JaqLsbpg38GwehL2dIS5OOfjoCuhE1dRW3LNQKAhUF2B/hYo+gQ2HPXILrQCIavUSfbHF7TycoxtCYQoBLDkyT9sEPYwfzfWixbTNxysrfYodA9xvtxQOnw6yc1jr/fJKagtSGtzRysdf+AtNztzq6cM58YfB5ABbOOd2564FVJk3V+oZeDW/RMrIuFJMGJubEwIWtqVmL0s63L2YlPohXC3A1umhFCBlILvEcrBLZ6PUoxsSCYvcc+DL9Xz82bUQ4nLwHs3mbHrwGMy9cNMYIjq6LihUbWMMuh27J3CnDHni8E1Dw6SyceYZm8yGoD12y7bx/X/yuw+uSYas4jePncVITCYsgOIRxN5QHBbwZ3TXmUT/yVpxHa2OkTjFuuNysEpgZ3COunydXyRRiajOQJlTXk4USt9ptXh6FecujawDhkm6NEJWeEMm+FeIe8PpF8iK5xBVbYPCOGTwEO+fb9Q/aB2UaDcO6ZKuTIu2Ys4qq/6VhMMxHxRJLGn87mVEN7RIt1rTYnJF/0Tuf8lvoCFVbOijOELWi0HqrfvP+P7qD705Zi7KpAP2A8h3Vompsjh72O/AapD3AuWz/cj1jfPVBv9kZynVblE20nebQyNcPlN6ne5U8H9yio06dFNpSJlD+GFk+D4QrVEmBomdyhyBZVqJn0+kUSBf2EdwxPlkImPQ6RZdTuXNyScoSku0kjtfBrXcIkwNTCuSxrKjH249JyRQFnhccWAXupOZDwzkigQJcyBT90jtez4aJ9tn0+RZFMcg3RVP4mYXUosNXgCaaOg==)

These are are `to` relevant values:

1. `'/features'`
2. `{ name: 'features' }`
3. `{ name: 'features', params: { id: undefined } }`
4. `{ name: 'features', params: { id: null } }`
5. `{ name: 'features', params: { id: '' } }`

All of them lead to the same path.

Clicking around between them, it seems that 2, 3 and 4 are considered equivalent, which makes sense. It's 5 that I found most interesting. It seems to behave like it's a child of 2, 3 and 4.

I also tried with `{ name: 'features', params: { id: '123' } }`, which seems to behave similarly. When that route is active it shows 2, 3 and 4 as active, though not exact active.

Not sure if that's intentional or not, but it doesn't seem to match what's described at <https://router.vuejs.org/guide/essentials/active-links.html>. I wrote that page, so I may be to blame for describing these classes incorrectly. It seems that there's some special handling for optional `params` that I didn't take into account in my description.

packages/router/__tests__/matcher/resolve.spec.ts
```diff
@@ -685,12 +685,31 @@ describe('RouterMatcher.resolve', () => {
...
+      )
+    })
+
+    // #2419: resolving by string path must match the same-named route's
+    // missing-optional-params behavior so active-link comparisons stay in sync
+    it('drops missing optional splat when resolving by path', () => {
+      assertRecordMatch(
+        { path: '/features/:pathMatch(.*)*', components, name: 'features' },
+        { path: '/features' },
+        { path: '/features', params: {}, name: 'features' }
```
posva, line 703: still unsure if params should be `{ pathMatch: [] }` for convenience since it's repeatable. This is probably not worth changing in the current version but worth changing in the upcoming route resolvers

posva (approved):

PR #2464, opened 2025-02-19
chore: build path compatibility with nodejs v20

SpanManX:
The expression path.dirname(new URL(import.meta.url).pathname) returns a path with a double G:\ prefix on nodejs v20, causing path resolution errors. Modify it using a ternary operator to make it compatible with nodejs v20.

### DIFF
--- a/packages/docs/run-typedoc.mjs
+++ b/packages/docs/run-typedoc.mjs
@@ -2,8 +2,7 @@ import path from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { createTypeDocApp } from './typedoc-markdown.mjs'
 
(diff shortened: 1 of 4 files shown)


posva (changes requested):
Thanks! I think we can make this so it doesn’t require __dirname anymore. The same code would probably work in all cases without win32 cases

SpanManX:
> Thanks! I think we can make this so it doesn’t require __dirname anymore. The same code would probably work in all cases without win32 cases

Change the code to "const __dirname = path.dirname(fileURLToPath(import.meta.url))".

skirtles-code:
@posva I'm unclear why this has been closed. The changes I made in #2466 were only intended to address the formatting issues in this PR, they don't fix the original problem.

posva:
I misunderstood it then. Sorry about that

PR #2465, opened 2025-02-19
fix: missing viewTransition prop on RouterLink

lubomirblazekcz:


posva (commented):
Thanks! Have you found any case where the lack of the js declaration has an impact?

lubomirblazekcz:
Yeah, when I added view-transition="true" on the router-link it didn't work for me at all. After this change it started working as expected

PR #2466, opened 2025-02-26
chore: change yorkie to simple-git-hooks

skirtles-code:
While reviewing #2464, I noticed that the formatting was incorrect on the imports. I wondered why that hadn't been automatically fixed on commit.

There are several reasons why this isn't working. I didn't want to create a huge PR, so here I've just switched `yorkie` to `simple-git-hooks`. By itself, this wouldn't have prevented the formatting problems seen in #2464, as it isn't checking `.mjs` files. Changing that is more complicated than it sounds, so I'll open other PRs to fix the other problems.

`yorkie` doesn't work with pnpm, the git hooks aren't created. For anyone who cloned the repo before it used pnpm, it might appear as though everything is working fine, but the git hooks aren't created for a fresh clone.

Most of the Vue ecosystem has moved away from `yorkie` and is now using `simple-git-hooks`, including Vue core and Pinia.

For reference, https://github.com/vuejs/core/pull/7069 is the PR that migrated Vue core. The changes I've made here are very similar.

Close #2464

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -34,9 +34,9 @@ Hi! I'm really excited that you are interested in contributing to Vue Router. Be
 
 - Make sure tests pass!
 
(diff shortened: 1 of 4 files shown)


posva (approved):
Thanks a lot! This is long overdue

PR #2467, opened 2025-02-26
chore: upgrade to Prettier 3

skirtles-code:
Apart from `package.json` and `pnpm-lock.yaml`, all the changes come from running `lint:fix`.

Prettier 2 doesn't support using `with { type: 'json' }` for imports, which we're using in some `.mjs` files. Currently we aren't linting `.mjs` files, but that's something I'd eventually like to change. Upgrading Prettier is a first step towards that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "lint-staged": "^15.2.10",
     "minimist": "^1.2.8",
     "p-series": "^3.0.0",
-    "prettier": "^2.8.8",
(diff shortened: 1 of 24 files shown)


posva (approved):
Thanks a lot!

posva:
In the next version we will likely upgrade to use type: module as well. The json import can always be rewritten as a read file + JSON.parse

PR #2470, opened 2025-03-04
chore: lint more files

skirtles-code:
There are several files that aren't currently being formatted by Prettier. I've tried to expand that to include more files.

- Add `mjs` and `mts` extensions.
- Add `*.html` to `lint-staged`.
- Expand the paths in `lint:script` to cover more files in `packages`, as well as `scripts`.

Changing the paths in `lint:script` is probably the trickiest part. It's not entirely clear to me whether I've got those paths correct.

It's worth noting that `lint-staged` doesn't take those paths into account, it just formats anything with the appropriate extensions. However, it only formats *staged* files, so anything in `.gitignore` is implicitly ignored by `lint-staged`.

Prettier 3 does take the root `.gitignore` into account, but it doesn't consider nested `.gitignore` files, which are being used in this project. I needed to add the VitePress cache to `.prettierignore` to account for that. Perhaps other files should be added too, but it wasn't clear to me what some of the things in the nested `.gitignore` files were.

There are also 3 VitePress config files in this PR with minor formatting tweaks. These are files that are now picked up by `lint:script` that weren't previously. They should have already been picked up by `lint-staged`, but prior to #2466 that wasn't working for most people.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,4 @@ __build__
 dist
 coverage
 tests_output
+packages/docs/.vitepress/cache
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "play": "pnpm run -r play",
     "build:size": "pnpm run -r build:size",
     "lint": "pnpm run lint:script && pnpm run lint:html",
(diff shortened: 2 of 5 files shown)


posva (approved):
Thanks!

PR #2475, opened 2025-03-13
feat: make typed `useRoute(currentRouteName)` return children route types as well

Anoesj:
In an app (with a typed router) where route `/a/b` is a child of route `/a`, this makes `useRoute('/a')` return `RouteLocationNormalizedLoadedTyped<RouteNamedMap, "/a"> | RouteLocationNormalizedLoadedTyped<RouteNamedMap, "/a/b">`.

It works in a recursive manner, so whenever route `/a/b/c` is added as a child of `/a/b`,  `useRoute('/a')` will return `RouteLocationNormalizedLoadedTyped<RouteNamedMap, "/a"> | RouteLocationNormalizedLoadedTyped<RouteNamedMap, "/a/b">` | RouteLocationNormalizedLoadedTyped<RouteNamedMap, "/a/b/c">`.

Of course `useRoute('/a/b/c')` will *only* return `RouteLocationNormalizedLoadedTyped<RouteNamedMap, "/a/b/c">`.

> [!NOTE] 
> This PR is intended to be paired with the following `unplugin-vue-router` PR: https://github.com/posva/unplugin-vue-router/pull/602
> That PR changes the DTS generation to include `RouteMeta` and a union of the children route's names as generics to every `RouteRecordInfo` type.

### DIFF
--- a/packages/router/src/typed-routes/route-map.ts
+++ b/packages/router/src/typed-routes/route-map.ts
@@ -18,7 +18,7 @@ export interface RouteRecordInfo<
   ParamsRaw extends RouteParamsRawGeneric = RouteParamsRawGeneric,
   Params extends RouteParamsGeneric = RouteParamsGeneric,
   Meta extends RouteMeta = RouteMeta,
-  _ChildrenRouteNames extends string | symbol = never,
+  ChildrenNames extends string | symbol | never = never | string | symbol,
 > {
   name: Name
   path: Path
@@ -26,6 +26,7 @@ export interface RouteRecordInfo<
   params: Params
   // TODO: implement meta with a defineRoute macro
   meta: Meta
+  childrenNames: ChildrenNames
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/router/src/typed-routes/route-map.ts
```diff
@@ -18,13 +18,15 @@ export interface RouteRecordInfo<
   ParamsRaw extends RouteParamsRawGeneric = RouteParamsRawGeneric,
   Params extends RouteParamsGeneric = RouteParamsGeneric,
   Meta extends RouteMeta = RouteMeta,
+  ChildrenNames extends string | symbol | never = never | string | symbol,
```
posva, line 21: Still figuring out a way to keep a RouteRecordInfo with type params assignable to the generic one (no type params).

posva (approved):
I think this is ready!

packages/router/src/typed-routes/route-map.ts
```diff
@@ -17,16 +13,30 @@ export interface RouteRecordInfo<
   // TODO: could probably be inferred from the Params
   ParamsRaw extends RouteParamsRawGeneric = RouteParamsRawGeneric,
   Params extends RouteParamsGeneric = RouteParamsGeneric,
-  Meta extends RouteMeta = RouteMeta,
```
posva, line 20: I removed this type param because it was never used and I figured it's safer to add it back later if needed than leaving an unused type param that can _eventually_ be used. I marked this as a fix because the _leftover_ type param should have never been released in the first place

posva (approved):

Issue #1517, opened 2022-08-14
Add `onBeforeRouteEnter` API to the list of composition API


ahnpnl:
### What problem does this feature solve?
Current the documentation about composition API https://router.vuejs.org/guide/advanced/navigation-guards.html#using-the-composition-api doesn't mention about `onBeforeRouteEnter`. I see in `Options API`, there is `beforeRouteEnter` hence there should be one for composition API too.

### What does the proposed API look like?
```
export function onBeforeRouteUpdate(to, from) {
      //....
}
```

posva:
`onBeforeRouteEnter()` doesn't exist, it's not possible to implement such guard with the composition api.

a798032825:
> `onBeforeRouteEnter()` doesn't exist, it's not possible to implement such guard with the composition api.

hello,may I get a case to put beforeRouteEnter into <script setup>.
just as onBeforeRouteleave can do.

719media:
@posva Can you help us understand a little more on this? I'm confused because beforeRouteEnter exists in options API here: https://router.vuejs.org/guide/advanced/navigation-guards.html#using-the-composition-api

Not sure why it then couldn't be made to be used inside of setup() ?

posva:
It’s a technical limitation: setup runs when the component is created which happens after the navigation. Therefore you cannot add a navigation guard while the component is created

taras-turchenko-moc:
@posva Hi! Can we do onBeforeRouteEnter hook using compiler macros? We can add this hook using `defineOptions` so I believe it's possible to make some shorthand for this to align options and composition apis

```js
defineOptions({
    beforeRouteEnter: preloadGuard(() => loadInitial())
});
```

Also it would be nice to add route meta from component like nuxt doing

posva:
The prefered direction are data loaders: https://uvr.esm.is/data-loaders/. You also have a `definePage()` macro for meta

PR #2490, opened 2025-04-14
chore(playground): upgrade to `script setup` syntax and TypeScript

curtgrimes:
Upgrade the playground package to use the `<script setup>` syntax and TypeScript. This change makes the playground more consistent with recent Vue.js practices and improves type safety.

Keep the purpose of each example identical to the original examples as much as possible. Therefore, components using `beforeRouteEnter()`, [which does not have a `<script setup>` equivalent](https://github.com/vuejs/router/issues/1517), are not converted to `<script setup>` syntax but are only converted to use TypeScript for improved type safety (Home.vue and ComponentWithData.vue). Converting these to use the [recommended data loaders or `defineMacro()`](https://github.com/vuejs/router/issues/1517#issuecomment-2263346429) could be an option in the future.

Update `RouteNamedMap` in main.ts to improve type safety and better demonstrate the newer typed routes feature (introduced in vue-router@4.4.0) in LongView.vue, Nested.vue, and RepeatedParams.vue.

Fix type issue in AppLink.vue where `RouterLinkProps` replaces `RouterLink.props`.

Remove the `name` option from the `defineComponent` calls that used it, because this option is [only useful when the component recursively calls itself](https://v2.vuejs.org/v2/api/#name), but there are not components in this playground project that do so.

### DIFF
--- a/packages/playground/src/AppLink.vue
+++ b/packages/playground/src/AppLink.vue
@@ -6,7 +6,7 @@
     :class="classes"
     target="_blank"
     rel="noopener noreferrer"
-    :href="to"
+    :href="String(to)"
     :tabindex="disabled ? -1 : undefined"
     :aria-disabled="disabled"
   >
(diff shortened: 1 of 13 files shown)


curtgrimes:
Sounds good! Thank you for the update.

posva (approved):
Thanks!

PR #2497, opened 2025-04-25
ci: avoid CI failures for forks

skirtles-code:
Some of the CI jobs only work for `vuejs/router` and fail when pushing changes to forks.

This PR restricts those jobs so that they don't run on forks.

The approach is based on:

- https://github.com/vuejs/core/blob/a23fb59e83c8b65b27eaa21964c8baa217ab0573/.github/workflows/ci.yml#L19

The existing check on `github.event.pull_request.head.repo.fork` doesn't seem to work for pushes, presumably because `github.event.pull_request` is only available for PRs and not pushes.

In theory, somebody might want to run these workflows on their fork, but I think that only makes sense if they don't intend to merge them back into the main repo, in which case they can just edit the workflow configs on their fork.

### DIFF
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -18,6 +18,7 @@ on:
 
 jobs:
   build:
+    if: github.repository == 'vuejs/router'
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #2498, opened 2025-05-02
feat: add data attributes for active links in RouterLink component

xerox0213:
## Summary

This PR enhances the `<RouterLink>` component by injecting two new data attributes on the rendered `<a>` element:

* `data-is-active` — present when the link matches the current route (inclusive match)
* `data-is-exact-active` — present when the link exactly matches the current route

These attributes allow authors to target active link states via attribute selectors (e.g. in Tailwind CSS) with higher specificity than relying solely on utility class ordering.

---

## Motivation

Tailwind’s JIT-generated CSS emits utility rules in a deterministic sequence. If the base link color rule happens to appear **after** your active-color rule in the compiled stylesheet, the base color will win and your active style won’t apply. By exposing `data-is-active` and `data-is-exact-active`, developers can write attribute-based selectors that reliably override base styles regardless of class order. For example:

```jsx
<RouterLink
  to="/posts"
  class="
    text-blue-800 
    hover:text-blue-600 
    data-[is-active=true]:text-blue-600 
  "
>
  Posts
</RouterLink>
```

### DIFF
--- a/packages/router/__tests__/RouterLink.spec.ts
+++ b/packages/router/__tests__/RouterLink.spec.ts
@@ -991,4 +991,39 @@ describe('RouterLink', () => {
       })
     })
   })
+  describe('data attributes for active links', () => {
+    it('sets data-is-active and data-is-exact-active when the link is exactly active', async () => {
+      const { wrapper } = await factory(
+        locations.basic.normalized,
+        { to: locations.basic.string },
+        locations.basic.normalized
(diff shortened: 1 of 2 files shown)


posva:
Thanks a lot for the PR. Unfortunately: https://github.com/vuejs/router/issues/2499#issuecomment-2847478143

PR #2502, opened 2025-05-21
chore(router): copy README file from workspace root

Kshao123:
add README to make the package look less bare on npm

> https://docs.npmjs.com/about-package-readme-files

### DIFF
--- a/packages/router/package.json
+++ b/packages/router/package.json
@@ -104,7 +104,9 @@
     "test:e2e:native": "node e2e/runner.mjs --env chrome",
     "test:e2e:ci": "node e2e/runner.mjs --env chrome-headless --retries 2",
     "test:e2e:bs": "node e2e/runner.mjs --local -e android5 --tag browserstack",
-    "test:e2e:bs-test": "node e2e/runner.mjs --local --env browserstack.local_chrome --tag browserstack"
+    "test:e2e:bs-test": "node e2e/runner.mjs --local --env browserstack.local_chrome --tag browserstack",
+    "prepack": "cp ../../README.md ./README.md",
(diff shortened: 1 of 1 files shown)


packages/router/package.json
```diff
@@ -104,7 +104,9 @@
     "test:e2e:native": "node e2e/runner.mjs --env chrome",
     "test:e2e:ci": "node e2e/runner.mjs --env chrome-headless --retries 2",
     "test:e2e:bs": "node e2e/runner.mjs --local -e android5 --tag browserstack",
-    "test:e2e:bs-test": "node e2e/runner.mjs --local --env browserstack.local_chrome --tag browserstack"
+    "test:e2e:bs-test": "node e2e/runner.mjs --local --env browserstack.local_chrome --tag browserstack",
+    "prepack": "cp ../../README.md ./README.md",
+    "postpack": "rm ./README.md"
```
skirtles-code, line 109: I don't think using `cp` and `rm` is portable. The package `rimraf` is already included, which is a portable version of `rm`. In the past I've used the `copyfiles` package as a portable replacement for `cp`.

skirtles-code (commented):
This used to be handled in `release.mjs`, but it seems that was removed in e03edfd9c47eff50607c1a7fee4d309fefab17c9. Unclear whether that was intentional. Maybe the plan was to handle this like Pinia, which uses a separate, shorter version of its `README.md` for the npm release?

Not sure which approach posva would prefer for Vue Router.

posva (changes requested):
@skirtles-code is right, this was an oversight when updating the release script. It should be moved there like it used to be

PR #2532, opened 2025-07-18
docs:  scroll-behavior.md  fix falsy to false

xl495:


posva:
This is correct

xl495:
> This is correct

ok, Thanks.

PR #2537, opened 2025-08-03
fix(router): use pagehide for iOS navigation

### LINKED ISSUE #2536: Scroll is not restored after a refresh on iOS / mobile safari
### Version
4.5.1

### Reproduction link
Any vueJS website (omnomnom.app is my project/you can try it out there as well/don't mean to advertise...)

### Steps to reproduce
Go to any vue website using vue-router, scroll to the middle of the page, refresh, it'll go back to top instead of going back to the correct position.

### What is expected?
It behaves properly in chrome for e.g., it'd restore the scroll position.

It's going back to the top of the page instead of restoring the scroll position.

---
The code window.addEventListener('beforeunload', beforeUnloadListener) does not trigger on iOS, we should use window.addEventListener('pagehide', ...) instead

omnomnomapp:
fix #2536

### Steps to reproduce
On iOS safari, scroll to the middle of the page, refresh, it'll go back to top instead of going back to the correct position.

### What is expected?
It behaves properly in chrome for e.g., it'd restore the scroll position.

It's going back to the top of the page instead of restoring the scroll position.

---
The code window.addEventListener('beforeunload', beforeUnloadListener) does not trigger on iOS, we should use window.addEventListener('pagehide', ...) instead

### DIFF
--- a/packages/router/src/history/html5.ts
+++ b/packages/router/src/history/html5.ts
@@ -126,9 +126,7 @@ function useHistoryListeners(
     return teardown
   }
 
-  function beforeHiddenListener() {
-    document.hidden && beforeUnloadListener()
-  }
+  function beforeHiddenListener() { document.hidden && beforeUnloadListener() }
 
   function beforeUnloadListener() {
     const { history } = window
@@ -143,7 +141,6 @@ function useHistoryListeners(
     for (const teardown of teardowns) teardown()
     teardowns = []
     window.removeEventListener('popstate', popStateHandler)
-    window.removeEventListener('beforeunload', beforeUnloadListener)
     window.removeEventListener('pagehide', beforeUnloadListener)
     document.removeEventListener('visibilitychange', beforeHiddenListener)
   }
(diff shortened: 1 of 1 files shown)


packages/router/src/history/html5.ts
```diff
@@ -140,15 +144,20 @@ function useHistoryListeners(
     teardowns = []
     window.removeEventListener('popstate', popStateHandler)
     window.removeEventListener('beforeunload', beforeUnloadListener)
```
posva: I think we can remove the beforeunload if we use pagehide

packages/router/src/history/html5.ts
```diff
@@ -126,6 +126,10 @@ function useHistoryListeners(
     return teardown
   }
 
+  function beforeHiddenListener() {
+    document.hidden && beforeUnloadListener()
```
posva: Inline this function instead

omnomnomapp:
Provided another commit to fix those, thanks for the fast feedback :)

PR #2547, opened 2025-09-08
docu: add missing export line in router.js code block

CodingNiclas:
This PR adds a missing export statement in the `router.js` code block of the documentation. 
Without this line, the code does not work properly.

## Summary by CodeRabbit

* **Documentation**
  * Updated routing guide examples to declare the router as a named export, making it easier to import by name in snippets.
  * Notes clarify this is a documentation-only change that doesn't affect runtime behavior or existing configurations.
  * Aligns examples with common import patterns to reduce confusion and improve developer experience.

### DIFF
--- a/packages/docs/guide/index.md
+++ b/packages/docs/guide/index.md
@@ -67,6 +67,8 @@ const router = createRouter({
   history: createMemoryHistory(),
   routes,
 })
+
+export default router
 ```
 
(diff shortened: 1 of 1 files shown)


packages/docs/guide/index.md
```diff
@@ -67,6 +67,8 @@ const router = createRouter({
   history: createMemoryHistory(),
   routes,
 })
+
+export default router
```
posva: Let's instead just add `export` in from of `const router` and do a named export

posva (approved):

PR #2617, opened 2026-02-07
docs: update Playground example

skirtles-code:
I've updated the Playground example to use the latest versions of `vue-router` and `@vue/devtools-api`:

image

This didn't require any code changes to the example, just changes to the `Import Map`.

I've also added a comment in `router.js`:

image

We occasionally get questions on Vue Land about routing being 'broken' and it seems some people just copy the code from that example without reading the accompanying explanation on the **Getting Started** page. This extra comment might to help to mitigate that confusion.

## Summary by CodeRabbit

* **Documentation**
  * Updated Vue Playground example links in guides to point to new target URLs.


posva (approved):
Nice, thanks!

PR #2625, opened 2026-02-13
docs: Fix route-map.d.ts path

bperel:
By default `route-map.d.ts` is stored in `src`, as shown in the `vite.config.ts` diff, so the `tsconfig.json` diff should reflect that as well.

## Summary by CodeRabbit

* **Documentation**
  * Updated the migration guide to clarify build-tool and TypeScript configuration, added an example showing how to generate and reference route type definitions, corrected include/path references and troubleshooting guidance to point to the generated types file. These changes help ensure generated types are referenced correctly during migration.

### DIFF
--- a/packages/docs/guide/migration/v4-to-v5.md
+++ b/packages/docs/guide/migration/v4-to-v5.md
@@ -149,7 +149,7 @@ or to your `tsconfig.json`:
   "include": [
     "./typed-router.d.ts", // [!code --]
     "unplugin-vue-router/client", // [!code --]
-    "./src/route-map.d.ts", // [!code ++]
+    // ...
   ],
 }
 ```


packages/docs/guide/migration/v4-to-v5.md
```diff
@@ -149,14 +149,14 @@ or to your `tsconfig.json`:
   "include": [
     "./typed-router.d.ts", // [!code --]
     "unplugin-vue-router/client", // [!code --]
-    "./route-map.d.ts", // [!code ++]
+    "./src/route-map.d.ts", // [!code ++]
```
posva: Actually let’s replace this with 
// ...
To symbolize other lines

bperel: What do you mean? Do you not want to list `"./src/route-map.d.ts"` as part of the includes anymore?

posva: Exactly

PR #2634, opened 2026-02-24
types(router): add | undefined to optional properties

YevheniiKotyrlo:
## Problem

vue-router's published types use `prop?: T` without explicit `| undefined` on optional interface properties. This works for the default TypeScript configuration, but fails under `exactOptionalPropertyTypes: true` — a flag that TypeScript 5.9's `tsc --init` now enables by default.

The most visible failure is the interface inheritance in the experimental route resolver where `EXPERIMENTAL_ResolverRecord_Group` narrows base properties to `?: undefined`, which requires `undefined` to be assignable to the base property type.

This also affects all other optional properties in published types (`RouteLocationOptions`, `_RouteRecordBase`, `RouteRecordSingleView`, `ScrollPositionCoordinates`, etc.) for consumers with `exactOptionalPropertyTypes: true` and `skipLibCheck: false`.

## Solution

Add `| undefined` to all optional properties across published interfaces and type definitions. This follows the exact same approach as vuejs/core#12771 ("types(jsx): add undefined to optional properties") which was merged for `@vue/runtime-dom`.

**No runtime changes. No tsconfig change. Only type annotations.**

### Not a breaking change

Per [semver-ts.org](https://www.semver-ts.org/formal-spec/5-compiler-considerations.html): changes which correctly represent runtime behavior are not breaking. Adding `| undefined` makes the types honest about what was always possible at runtime.

- `exactOptionalPropertyTypes: false` (default): **zero impact** — types are identical
- `exactOptionalPropertyTypes: true`: **fixes TS2375 errors**

### Scope: 8 files

- `src/types/index.ts` — `RouteQueryAndHash`, `RouteLocationOptions`, `LocationAsRelativeRaw`, `MatcherLocationAsName`, `MatcherLocationAsRelative`, `_RouteRecordBase`, all `RouteRecordSingleView`/`MultipleViews`/`Redirect` variants
- `src/typed-routes/route-location.ts` — `RouteLocationAsRelativeGeneric`, `RouteLocationAsRelativeTyped`
- `src/scrollBehavior.ts` — `ScrollPositionCoordinates`, `_ScrollPositionNormalized` (type definitions only)
- `src/matcher/pathParserRanker.ts` — `_PathParserOptions`
- `src/experimental/route-resolver/resolver-fixed.ts` — `EXPERIMENTAL_ResolverRecord_Base`
- `src/experimental/route-resolver/resolver-abstract.ts` — `ResolverLocationAsNamed`, `ResolverLocationAsPathRelative`, `ResolverLocationAsRelative`
- `src/experimental/router.ts` — `EXPERIMENTAL_RouterOptions_Base`, `EXPERIMENTAL_RouteRecord_Base`, subrecords
- `src/unplugin/options.ts` — `Options`, `RoutesFolderOption`, `ParamParsersOptions`

## Tests

- `pnpm -C packages/router run build` — ✅
- `pnpm -C packages/router run test:types` — 0 errors
- `pnpm -C packages/router run test:unit` — 1494 passed

### DIFF
--- a/packages/router/src/scrollBehavior.ts
+++ b/packages/router/src/scrollBehavior.ts
@@ -12,9 +12,9 @@ import { warn } from './warning'
  * Note that not all browsers support `behavior`.
  */
 export type ScrollPositionCoordinates = {
-  behavior?: ScrollOptions['behavior']
-  left?: number
-  top?: number
+  behavior?: ScrollOptions['behavior'] | undefined
+  left?: number | undefined
+  top?: number | undefined
 }
 
 /**
@@ -24,7 +24,7 @@ export type ScrollPositionCoordinates = {
  * @internal
  */
 export type _ScrollPositionNormalized = {
-  behavior?: ScrollOptions['behavior']
+  behavior?: ScrollOptions['behavior'] | undefined
   left: number
   top: number
 }
--- a/packages/router/src/experimental/route-resolver/resolver-abstract.ts
+++ b/packages/router/src/experimental/route-resolver/resolver-abstract.ts
@@ -147,8 +147,8 @@ export interface ResolverLocationAsNamed {
   name: RecordName
   // FIXME: should this be optional?
   params: MatcherParamsFormatted
-  query?: LocationQueryRaw
-  hash?: string
+  query?: LocationQueryRaw | undefined
+  hash?: string | undefined
 
   /**
    * @deprecated This is ignored when `name` is provided
@@ -168,8 +168,8 @@ export interface ResolverLocationAsNamed {
  */
 export interface ResolverLocationAsPathRelative {
   path: string
-  query?: LocationQueryRaw
-  hash?: string
+  query?: LocationQueryRaw | undefined
+  hash?: string | undefined
 
   /**
    * @deprecated This is ignored when `path` is provided
@@ -208,9 +208,9 @@ export interface ResolverLocationAsPathAbsolute extends ResolverLocationAsPathRe
  * ```
  */
 export interface ResolverLocationAsRelative {
-  params?: MatcherParamsFormatted
-  query?: LocationQueryRaw
-  hash?: string
+  params?: MatcherParamsFormatted | undefined
+  query?: LocationQueryRaw | undefined
+  hash?: string | undefined
 
   /**
    * @deprecated This location is relative to the next parameter. This `name` will be ignored.
--- a/packages/router/src/experimental/route-resolver/resolver-fixed.ts
+++ b/packages/router/src/experimental/route-resolver/resolver-fixed.ts
@@ -32,35 +32,35 @@ export interface EXPERIMENTAL_ResolverRecord_Base {
    * Name of the matcher. Unique across all matchers. If missing, this record
    * cannot be matched. This is useful for grouping records.
    */
-  name?: RecordName
+  name?: RecordName | undefined
 
   /**
    * {@link MatcherPattern} for the path section of the URI.
    */
-  path?: MatcherPatternPath
+  path?: MatcherPatternPath | undefined
 
   /**
    * {@link MatcherPattern} for the query section of the URI.
    */
-  query?: MatcherPatternQuery[]
+  query?: MatcherPatternQuery[] | undefined
 
   /**
    * {@link MatcherPattern} for the hash section of the URI.
    */
-  hash?: MatcherPatternHash
+  hash?: MatcherPatternHash | undefined
 
   /**
    * Parent record. The parent can be a group or a matchable record.
    * It will be included in the `matched` array of a resolved location.
    */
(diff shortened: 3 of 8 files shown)


packages/router/src/experimental/data-loaders/auto-exports.ts
```diff
@@ -76,6 +76,8 @@ export function AutoExportLoaders({
     transform: {
       order: 'post',
       filter: {
+        // @ts-expect-error: unplugin's StringFilter adds `| undefined` to optional include/exclude
+        // but vite's StringFilter omits it; structurally compatible at runtime
         id: transformFilter,
```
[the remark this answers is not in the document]
YevheniiKotyrlo: The `@ts-expect-error` is intentionally self-documenting here — if the upstream type gap closes, the directive will naturally fail the build and force its removal. That's the advantage of `@ts-expect-error` over `@ts-ignore`. The existing comment already explains the cross-library mismatch. No FIXME needed.

posva:
Thanks a lot! I will have to take a proper look, but ideally there should be no runtime changes. The `meta` property shouldn't allow `undefined` as a value, that would make the spread work

packages/router/__tests__/guards/onBeforeRouteUpdate.spec.ts
```diff
@@ -27,7 +27,7 @@ const component = {
 function withSpy(name?: string, isAsync = false) {
   const spy = vi.fn()
   const Component = defineComponent({
-    name,
+    ...(name != null ? { name } : {}),
```
posva: Shouldn't this work in Vue?

In this case, I think it's better to have the fallback

```suggestion
    name: name || 'no-name',
```

packages/router/src/matcher/index.ts
```diff
@@ -377,24 +377,36 @@ export function createRouterMatcher(
...
+    if (key in params) {
+      const value = params[key]
+      if (value !== undefined) newParams[key] = value
+    }
   }
 
   return newParams
 }
 
+function toPathParams(params: MatcherLocation['params']): PathParams {
```
posva: This change to the matcher shouldn't be runtime

packages/router/src/scrollBehavior.ts
```diff
@@ -81,7 +81,7 @@ function getElementPosition(
   const elRect = el.getBoundingClientRect()
 
   return {
-    behavior: offset.behavior,
+    ...(offset.behavior != null && { behavior: offset.behavior }),
```
posva: this one shouldn't be needed either

packages/router/src/RouterLink.ts
posva: shouldn't be needed either, the type should be stricter

posva (changes requested):
A few other places where the runtime is still affected

packages/router/src/experimental/data-loaders/defineColadaLoader.ts
```diff
@@ -44,6 +44,7 @@ import type {
 import { useRoute, useRouter } from '../../useApi'
 import type { Router } from '../../router'
 import type { LocationQuery } from '../../query'
+import type { RouteParamsGeneric } from '../../types'
```
posva: let's revert this, it was intentional

packages/router/src/experimental/data-loaders/defineColadaLoader.ts
```diff
@@ -732,7 +733,7 @@ export interface DataLoaderColadaEntry<
 
 interface TrackedRoute {
   ready: boolean
-  params: Partial<LocationQuery>
+  params: Partial<RouteParamsGeneric>
```
posva: same here

packages/router/src/experimental/data-loaders/utils.ts
```diff
@@ -113,10 +113,10 @@ function trackObjectReads<T extends Record<string, unknown>>(obj: T) {
  * @param outer - the bigger params
  * @param inner - the smaller params
  */
-export function isSubsetOf(
-  inner: Partial<LocationQuery>,
-  outer: LocationQuery
-): boolean {
+export function isSubsetOf<
```
posva: No need for the extra type complexity here

posva (changes requested):

YevheniiKotyrlo:
Thanks for the follow-up review. You're right — the `defineColadaLoader` and `utils` changes went beyond what this PR should be doing.

I've stripped this down to strictly `| undefined` on published type annotations. Zero runtime changes, zero internal logic changes, zero test changes. Single commit on latest main.

The `RouteParamsGeneric` value type widening and any internal fixes needed for enabling `exactOptionalPropertyTypes` in the router's own tsconfig can be discussed separately as a follow-up — this PR is just about making the published types compatible for consumers who enable the flag.

Issue #1780, opened 2023-04-06
Allow passing transient `info` in navigations


atom-git:
### What problem is this solving

This would allow passing information specific to the navigation (https://github.com/WICG/navigation-api#example-using-info)

### Proposed solution

TBD
This requires an RFC

posva:
This is what we call transient state (state available only during navigation) and is planned in a way that would align with the upcoming navigation API (https://github.com/WICG/navigation-api#example-using-info). It would require going through an RFC first though

jods4:
Big +1 for transient state!
- It has some useful applications (not the least being force reloading a component when the route doesn't change);
- and it's incredibly difficult to work around, the router code goes to great lengths to "normalize" everything passed in the route and only selectively copies properties.

There are few places where you could pass "hidden" state, most likely `params`, but in most cases only known keys will passthrough. I say "in most cases" because the code around this is relatively inconsistent. For example if you're modifying the current route and pass no `name`, `params` doesn't get normalized (for unclear reasons, seems pretty inconsistent).

d-ryabtsev:
We need this!

mhDuke:
## the goal:
> to attach some volatile state to the next route, this state should not be easily accessible by the end-user.

***navigation request***: the object we pass to `router.push(navigation-request)` or `router.replace()`. which is usually consists of a route `name` or route `path`, as well as optional `params`, and optional `query` objects.

there are couple suggestions:
## Through the navigation request `meta`.
for example
```js
const router = useRouter();
router.push({ 
  name: 'some-route-name', 
  meta: { pieceOfInfo: 'info' } 
});
```
then we match the route as usual by name or path, and then we have two options:
- we either override the meta field with whatever is given by the navigation request, or.
- we append (merge) what is given from the navigation request meta into the predefined route record meta, which is somewhat tricky.

## Through the navigation request new _reserved_ field, let's call transientState.
`transientState` is not a field that we can define on the route records themselves, rather, it's a field that can be passed in the navigation request object, beside name, path, query, params.
```js
const router = useRouter();
router.push({ 
  name: 'some-route-name', 
  transientState: 'this state will be accessible via the route object after the matching is successful'
});
```

this doesn't have to be aligned with other specifications, for example [the suggested navigation API](https://github.com/WICG/navigation-api), because our proposal here is only concerned with passing volatile state that cannot be captured by the browsers history API.

[...]

PR #2646, opened 2026-02-27
fix(unplugin): generate param types from override paths and stop inheritance on absolute overrides

G100my:
I discovered this problem while migrating from `vite-plugin-pages`.

### Problem

When overriding route `path` via `<route>` or `definePage()`:

1. Params from the override path are not always reflected in generated types.
2. Absolute override paths (`/`-prefixed) can still inherit ancestor params/query in types, producing invalid keys.

### Solution

1. Infer path params directly from override `path`, including:
   - optional: `:id?`
   - repeatable: `:id+` / `:id*`
   - splat: `:id(.*)`
2. Apply `params.path` parser overrides to inferred params.
3. Treat absolute override paths as inheritance boundaries:
   - stop inheriting ancestor params/query above that node.
4. Keep relative override paths inheriting parent params/query as-is.

## Summary by CodeRabbit

* **Bug Fixes**
  * Absolute path overrides now correctly exclude parent parameters, while relative overrides continue to inherit them.
  * Improved handling of multiple, optional, repeatable, query, and splat parameters in route paths.
  * Route matching now correctly identifies trailing splat segments, including nested routes.
  * Added a build-time warning when the same parameter parser is declared in both the filename and route configuration.

* **Tests**
  * Expanded coverage for path overrides, parameter inheritance, parser metadata, and splat behavior.

### DIFF
--- a/packages/router/src/unplugin/core/tree.spec.ts
+++ b/packages/router/src/unplugin/core/tree.spec.ts
@@ -727,6 +727,141 @@ describe('Tree', () => {
     } satisfies Partial<TreePathParam>)
   })
 
+  describe('path override and params extraction', () => {
+    it('reproduces missing params from custom path override', () => {
+      const tree = new PrefixTree(RESOLVED_OPTIONS)
+      const node = tree.insert('users/profile', 'users/profile.vue')
+
+      node.setCustomRouteBlock('users/profile.vue', {
+        path: '/users/:id',
+      })
+
+      expect(node.pathParams.map(param => param.paramName)).toEqual(['id'])
+    })
+
+    it('preserves parser, optional, repeatable, and splat metadata from custom path override', () => {
+      const tree = new PrefixTree(RESOLVED_OPTIONS)
+      const node = tree.insert('x/y', 'x/y.vue')
+      node.setCustomRouteBlock('x/y.vue', {
+        path: '/docs/:chapters+/:tags*/:path(.*)',
+        params: {
+          path: {
+            chapters: 'int',
+          },
+        },
+      })
+
+      expect(node.pathParams).toEqual([
+        expect.objectContaining({
+          paramName: 'chapters',
+          parser: 'int',
+          modifier: '+',
+          optional: false,
+          repeatable: true,
+          isSplat: false,
+        }),
+        expect.objectContaining({
+          paramName: 'tags',
+          modifier: '*',
+          optional: true,
+          repeatable: true,
+          isSplat: false,
+        }),
+        expect.objectContaining({
+          paramName: 'path',
+          modifier: '',
+          optional: false,
+          repeatable: false,
+          isSplat: true,
+        }),
+      ])
+    })
+
+    it('uses absolute path overrides as param inheritance boundaries', () => {
+      const tree = new PrefixTree(RESOLVED_OPTIONS)
+
+      const parent = tree.insert('org/[orgId]', 'org/[orgId].vue')
+      parent.setCustomRouteBlock('org/[orgId].vue', {
+        params: {
+          query: {
+            q: {},
+          },
+        },
+      })
+
+      const absoluteChild = tree.insert(
+        'org/[orgId]/dashboard',
+        'org/[orgId]/dashboard.vue'
+      )
+      absoluteChild.setCustomRouteBlock('org/[orgId]/dashboard.vue', {
+        path: '/dash/:id',
+        params: {
+          query: {
+            tab: {},
+          },
+        },
+      })
+
+      const relativeChild = tree.insert(
+        'org/[orgId]/reports',
+        'org/[orgId]/reports.vue'
+      )
+      relativeChild.setCustomRouteBlock('org/[orgId]/reports.vue', {
+        path: ':id',
+        params: {
+          query: {
+            tab: {},
+          },
+        },
+      })
+
+      const middle = tree.insert(
+        'org/[orgId]/projects/[projectId]',
+        'org/[orgId]/projects/[projectId].vue'
+      )
+      middle.setCustomRouteBlock('org/[orgId]/projects/[projectId].vue', {
+        path: '/p/:projectId',
+        params: {
+          query: {
+            page: {},
+          },
+        },
+      })
+
+      const leaf = tree.insert(
+        'org/[orgId]/projects/[projectId]/settings/[section]',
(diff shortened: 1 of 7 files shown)


packages/router/src/unplugin/core/tree.spec.ts
```diff
@@ -721,6 +721,182 @@ describe('Tree', () => {
     } satisfies Partial<TreePathParam>)
   })
 
+  describe('path override and params extraction', () => {
```
posva, line 730: I copy pasted these tests in main and they all pass. Could you provide one failing test?

posva (changes requested):

packages/router/src/unplugin/core/tree.spec.ts
G100my: I also tried copy-pasting the tests directly onto main, but they fail on my end.

I force-pushed an updated version that splits the changes into two commits: 
- the first one only adds the tests, 
- and the second one contains the actual code changes. 

Hopefully this makes it easier to verify without having to copy and paste the tests manually.
I also reworked the tests a bit to make them more focused.

posva (approved):
Thanks! There are a few things I found while working on this that I will fix locally after merging

posva:
@G100my what were the actual filepath and overrides you had in your app? I'm realizing having any kind of override for paths is too dangerous, especially absolute overrides. Knowing what you are trying to do will help me out too

enpitsuLin:
Hey @G100my, thanks for the fix — the new types do match runtime, no complaints there. 

***
cc @posva

The thing is that I've tried building a modal as a next.js-style parallel/intercepting route ([ref](https://nextjs.org/docs/app/api-reference/file-conventions/parallel-routes#modals)). An overlay route nested under my session layout (`[[id]]`) overrides its path to absolute `/settings`, and a plain full-page route uses the same URL. 

> The nesting is intentional — the session layout renders behind the dialog so the backdrop shows the current conversation (FWIW this also trips the VUE_ROUTER_R0104 warning at runtime which I ignored). In-app nav shows the overlay; hard loads fall back to the full page.

Found this by accident btw — while testing the pkg.pr.new preview of #2792, the generated `typed-router.d.ts` differed from 5.2.0: the override children's params went from `{ id?: ... }` to `Record<never, never>`. 

The issue: `useRoute<Name>()` is typed as `Name | RouteMap[Name]["childrenNames"]`, so in the session layout, route.params is the union of the layout + all the override children — and since some union members have no id, route.params.id no longer typechecks, even though the layout's own route still declares it (as optional):

```ts
const route = useRoute('session')
route.params.id // TS2339 — 'id' does not exist on
// 'Record<never, never> | { id?: string | undefined; } | { channelId: string; } | { providerId: string; }'
```

This type error breaks the build — any type-safe way to read the layout's own params from useRoute(name) here, or an intended pattern for this?

Happy to throw together a minimal repro if that helps. 

<details>
<summary>Actual file layout and path overrides</summary>

```
src/pages/
├── [[id]].vue                      # session layout — definePage({ name: 'session' }), path /:id?
├── [[id]]/settings.overlay.vue     # definePage({ path: '/settings', name: 'settings-overlay',
│                                   #   redirect: { name: 'settings-overlay-agent' }, meta: { overlay: true } })
├── [[id]]/settings.overlay/
│   ├── about.vue                   # definePage({ name: 'settings-overlay-about' })
│   ├── agent.vue                   # definePage({ name: 'settings-overlay-agent' })
│   ├── appearance.vue              # definePage({ name: 'settings-overlay-appearance' })
│   ├── channels.vue                # definePage({ name: 'settings-overlay-channels', redirect: ... })
│   ├── channels/index.vue          # definePage({ name: 'settings-overlay-channels-list' })
│   ├── channels/[channelId].vue    # definePage({ path: '/settings/channels/:channelId', name: 'settings-overlay-channels-detail' })
│   ├── compaction.vue              # definePage({ name: 'settings-overlay-compaction' })
│   ├── providers.vue               # definePage({ name: 'settings-overlay-providers', redirect: ... })
│   ├── providers/index.vue         # definePage({ name: 'settings-overlay-providers-list' })
│   └── providers/[providerId].vue  # definePage({ path: '/settings/providers/:providerId', name: 'settings-overlay-providers-detail' })
└── settings.vue                    # full-page counterpart — definePage({ redirect: { name: 'settings-agent' } }), path /settings
    └── about.vue …                 # settings-about, settings-agent, … (no overrides)
```

```ts
// src/router/index.ts — hard-load fallback
router.beforeEach((to, from) => {
  if (to.meta.overlay && from.matched.length === 0 && typeof to.name === 'string') {
    const fullPageName = to.name === 'settings-overlay'
      ? '/settings'
      : to.name.replace('settings-overlay', 'settings')
    return { name: fullPageName } as RouteLocationRaw
  }
})
```
</details>

posva:
The typesafe equivalent is

```ts
if ('id' in route) {
  route.id
}
```

In any case, I will come back to #2791 after #1780 with a better solution

G100my:
Hi, @posva

In my previous use case — which was actually only earlier in 2026... things are changing really fast 😅 — there were cases where certain pages needed to have specific, stable paths so that other services could link to them directly.

At the same time, within the file structure, we wanted the development team to be able to identify the purpose of different pages more easily, so we preferred using more descriptive filenames instead of structures like `xx/xx.xx/[id].vue`, where the page's purpose is not immediately obvious from the filename alone.

That was the main reason I used path overrides: to keep the externally exposed URL structure separate from the internal file naming and organization.

Looking back now, though, with the rapid development of AI-assisted tooling, engineers may gradually become less dependent on file names and directory structures for understanding a codebase. So adding complexity to routing overrides purely for the sake of file naming readability may not be worth the trade-off anymore.

If absolute path overrides make the package less safe or predictable, or add unnecessary complexity to the core logic, I'd support removing them in favor of keeping the base package behavior simpler and more consistent.

G100my:
And lastly, thank you for all the work you've put into the Vue ecosystem over the years. I really appreciate it.

posva:
@G100my thanks for the info. If it helps, you can (and IMO should) name params more explicitly like `[documentId]` and you can use groups to give more explicit names without affecting the url: `(dashboard)/controls.vue`.

One of the advantages of file-based approach is that collisions are harder (groups can still produce collisions). If we start overriding paths, then we need to read each file to know but with a file-based approach you only need to list files to know

PR #2661, opened 2026-03-16
fix(query): use Object.create(null) to prevent prototype pollution

### LINKED ISSUE #2658: parseQuery creates objects with `{}` instead of `Object.create(null)`, allowing `__proto__` key corruption
### Reproduction

Any Vue Router app using the default `parseQuery` implementation. Minimal test case:

```js
import { parseQuery } from 'vue-router'
const q = parseQuery('__proto__=evil&foo=bar')
console.log(q)           // unexpected ghost keys from prototype chain
console.log(q.__proto__)  // 'evil' shadows Object.prototype
```

1. Create a Vue app with Vue Router (or Nuxt, which uses Vue Router's default `parseQuery`)
2. Send a request with `__proto__` or `constructor` as a query parameter:
   ```bash
   curl 'http://localhost:3000/?__proto__=evil&foo=bar'
   ```
3. Observe the rendered `route.fullPath` — it contains ghost parameters:
   ```
   /?foo=bar&0=[object+Object]&1=evil
   ```
4. The `constructor` variant also leaks internal references:
   ```bash
   curl 'http://localhost:3000/?constructor=test'
   # Renders: /?constructor=function+Object()+{+[native+code]+}&constructor=test
   ```

### Expected behavior

`parseQuery` should return a null-prototype object (via `Object.create(null)`) so that inherited properties like `__proto__` and `constructor` cannot be accidentally read or overwritten. Query keys like `__proto__` should be stored as plain data without affecting the object's prototype chain.

### Actual behavior

`parseQuery` creates query objects with `const query = {}`. When `?__proto__=evil` is parsed, `"__proto__" in query` evaluates to `true` (inherited property), and the subsequent `query[key] = [currentValue]` corrupts the object. This causes `stringifyQuery` to emit ghost parameters from the corrupted prototype chain.

The corruption is scoped to that specific object instance — it does not achieve persistent cross-request `Object.prototype` pollution. But it does cause unexpected query parameters to appear in `route.fullPath`, which could interfere with middleware or logic that inspects `route.query`.

### Additional information

The root cause is in `parseQuery` (around line 1433-1449 in `vue-router.cjs`). Changing `const query = {}` to `const query = Object.create(null)` would fix this. Libraries like `ufo` already use `Object.create(null)` in their `parseQuery` implementation for this reason.

wdskuki:
## Summary

This PR fixes a security vulnerability where query objects created with `{}` are susceptible to prototype pollution via `__proto__` and `constructor` keys.

## Changes

- `parseQuery()`: Use `Object.create(null)` instead of `{}`
- `normalizeQuery()`: Use `Object.create(null)` instead of `{}`
- `parseURL()`: Use `Object.create(null)` for initial query object

## Security Impact

Before this fix, an attacker could craft a URL like:
- `/?__proto__=evil&foo=bar` - pollutes the query object's prototype
- `/?constructor=test` - leaks internal Object constructor

After this fix, these keys are treated as plain data without affecting the prototype chain.

## Fixes

Fixes #2658

## Summary by CodeRabbit

* **New Features**
  * Experimental query parsing and normalization added for more consistent query handling and URL normalization.
* **Bug Fixes**
  * Improved handling of repeated params, decoding edge cases, relative path/query/hash resolution, and safety against prototype-pollution-like issues.
* **Tests**
  * Expanded test coverage for query parsing, normalization, URL parsing, and resolver behavior.
* **Refactor**
  * Internal parsing wiring updated to use the new experimental implementations.

### DIFF
--- a/packages/router/src/experimental/data-loaders/defineColadaLoader.ts
+++ b/packages/router/src/experimental/data-loaders/defineColadaLoader.ts
@@ -43,7 +43,7 @@ import type {
 } from '../../typed-routes'
 import { useRoute, useRouter } from '../../useApi'
 import type { Router } from '../../router'
-import type { LocationQuery } from '../../query'
+import type { LocationQuery } from '../query'
 
 /**
  * Creates a Pinia Colada data loader with `data` is always defined.
--- a/packages/router/src/experimental/data-loaders/utils.ts
+++ b/packages/router/src/experimental/data-loaders/utils.ts
@@ -1,4 +1,4 @@
-import type { LocationQuery } from '../../query'
+import type { LocationQuery } from '../query'
 import type { Router } from '../../router'
 import type { RouteLocationNormalizedLoaded } from '../../typed-routes'
 import type { DataLoaderEntryBase, UseDataLoader } from './createDataLoader'
--- a/packages/router/src/experimental/location.spec.ts
+++ b/packages/router/src/experimental/location.spec.ts
@@ -1,0 +1,244 @@
+import { experimental_parseQuery as parseQuery } from './query'
+import { experimental_parseURL as originalParseURL } from './location'
+import { vi, describe, expect, it } from 'vitest'
+
+describe('parseURL', () => {
+  let parseURL = originalParseURL.bind(null, parseQuery)
+
+  it('works with no query no hash', () => {
+    expect(parseURL('/foo')).toEqual({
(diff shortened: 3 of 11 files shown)


posva (approved):
Thanks! Since it's a breaking change and the security impact doesn't seem high, I applied the change to the upcoming experimental router (where I can release breaking changes freely)

PR #2662, opened 2026-03-16
fix(matcher): add workaround hint to invalid params warning

### LINKED ISSUE #1617: Warning of invalid param for catch all route with named route redirect
### Version
4.1.6

### Reproduction link
[jsfiddle.net/z0kytrqj/16/](https://jsfiddle.net/z0kytrqj/16/)

### Steps to reproduce
Navigate to a catch all route with a named route redirect
```ts
  {
    path: '/:pathMatch(.*)*',
    redirect: { name: 'HOME' }
  }
```

### What is expected?
No warnings

Warning displayed: Discarded invalid param(s) "pathMatch" when navigating. See https://github.com/vuejs/router/blob/main/packages/router/CHANGELOG.md#414-2022-08-22 for more details.

shanliuling:
## What is the problem?

When using a catch-all route with a named redirect like:

```ts
{
  path: '/:pathMatch(.*)*',
  redirect: { name: 'HOME' }
}
```

Users see a warning:
`Discarded invalid param(s) "pathMatch" when navigating.`

The warning message doesn't explain how to fix the issue.

## How did I fix it?

Enhanced the warning message to include a helpful workaround hint:
`If you are using a catch-all route with a named redirect, pass an empty params object: redirect: { name: '...', params: {} }.`

This matches the workaround suggested by @posva in the issue comments.

## How was this tested?

- Updated the existing test in `packages/router/__tests__/warnings.spec.ts` to match the new warning format
- The test now also checks for the presence of the workaround hint

Closes #1617

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified discarded/invalid route-parameter warning: an extra guidance hint is now appended only for catch-all style routes with named redirects when the previous location contains params; other cases retain the original wording and link.

* **Tests**
  * Added and updated tests to verify the exact warning text and the conditional presence or absence of the workaround hint.

### DIFF
--- a/packages/router/__tests__/warnings.spec.ts
+++ b/packages/router/__tests__/warnings.spec.ts
@@ -314,7 +314,58 @@ describe('warnings', () => {
       }
     )
     expect('invalid param(s) "no", "foo" ').toHaveBeenWarned()
+    // workaround hint should NOT be shown for normal routes
+    expect('catch-all route with a named redirect').not.toHaveBeenWarned()
     // from the previous location
     expect('"one"').not.toHaveBeenWarned()
+  })
+
+  it('warns with workaround hint for catch-all route with named redirect', () => {
+    // This is the actual scenario from issue #1617
+    const record = {
+      path: '/:pathMatch(.*)*',
+      name: 'catch-all',
+      redirect: { name: 'HOME' }, // named redirect
+      components: {},
+    }
+    const matcher = createRouterMatcher([record], {})
+    // Use invalid params (foo, bar) to trigger the warning
+    // pathMatch is valid for catch-all, but foo/bar are not
+    matcher.resolve(
+      { name: 'catch-all', params: { foo: 'bar' } },
+      {
+        path: '/parent/one',
+        name: undefined,
+        params: { old: 'one' }, // has params
+        matched: [] as any,
+        meta: {},
+      }
+    )
+    expect('Discarded invalid param(s)').toHaveBeenWarned()
+    // workaround hint SHOULD be shown for catch-all with named redirect
+    expect('catch-all route with a named redirect').toHaveBeenWarned()
+  })
+
+  it('does not show hint when current location has no params', () => {
+    const record = {
+      path: '/:pathMatch(.*)*',
+      name: 'catch-all',
+      redirect: { name: 'HOME' },
+      components: {},
+    }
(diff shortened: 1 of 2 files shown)


posva (changes requested):
Thanks but this adds noise for all misses. It should only be displayed if the route is a named redirect with no params and the initial target location contained params

packages/router/__tests__/warnings.spec.ts
```diff
@@ -313,7 +313,9 @@ describe('warnings', () => {
         meta: {},
       }
     )
-    expect('invalid param(s) "no", "foo" ').toHaveBeenWarned()
+    expect('invalid param(s) "no", "foo"').toHaveBeenWarned()
+    // workaround hint should be shown
+    expect('If you are using a catch-all route').toHaveBeenWarned()
```
posva: shouldn't this not be shown? It would be nice to also have a dedicated unit test of the warning that is being added

posva (changes requested):

posva (changes requested):
The tests still do not look good: they don't showcase the redirect issue and the warning seems to appear even without it

packages/router/src/matcher/index.ts
```diff
@@ -263,10 +263,20 @@ export function createRouterMatcher(
         ).filter(paramName => !matcher!.keys.find(k => k.name === paramName))
 
         if (invalidParams.length) {
+          // Check if this is a catch-all route (contains pathMatch) with a named redirect
+          const isCatchAllWithRedirect =
+            matcher!.record.path.includes('pathMatch') && // catch-all route
```
posva: It’s not about the shape of the path, you need to check for params being passed implicitly.

posva (changes requested):

shanliuling:
Hi @posva, thanks for pointing out the implicit params logic! You were completely right—checking the shape of the path was fundamentally flawed.

I have completely refactored the approach:
1. Removed all \pathMatch\ and \
edirect\ shape checks.
2. The warning hint is now **only** triggered when the discarded parameters were **passed implicitly** (by verifying if \invalidParams\ exist in \currentLocation.params\), and the target route has no params.
3. Rewrote the tests to correctly showcase this exact implicit-passing scenario without adding noise to explicit misses.

This should be fully aligned with your intended logic. Please take a look!

PR #2683, opened 2026-04-13
fix(router): filter inherited params on named route redirect

### LINKED ISSUE #1617: Warning of invalid param for catch all route with named route redirect

mvtandas:
## Summary

Fixes #1617

When a catch-all route (e.g. `/:pathMatch(.*)*`) redirects to a named route via `redirect: { name: 'Home' }`, the `pathMatch` param from the source route was being passed to the target route. Since the target route doesn't accept that param, a spurious "Discarded invalid param(s)" warning was shown.

## Root Cause

In `handleRedirectRecord()`, when redirecting to a named route without an explicit `path`, all params from the source route (`to.params`) were passed through to the redirect target:

```ts
params: newTargetLocation.path != null ? {} : to.params,
```

This meant catch-all params like `pathMatch` were forwarded to the target named route, triggering the warning in `matcher.resolve()`.

## Fix

Added a `filterParams()` helper that uses `matcher.getRecordMatcher()` to look up the target route's accepted param keys and filters out any params that don't belong to the target route. This is only applied when the redirect target is a named route.

## Test

Added a test case that navigates to a catch-all route with a named redirect and verifies:
- Navigation succeeds to the correct target route
- No "invalid param" warning is emitted

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced redirect parameter handling for named route targets to prevent invalid parameter warnings.

* **Tests**
  * Added test coverage for named route redirect scenarios.

### DIFF
--- a/packages/router/__tests__/router.spec.ts
+++ b/packages/router/__tests__/router.spec.ts
@@ -77,6 +77,10 @@ const routes: RouteRecordRaw[] = [
     ],
   },
   { path: '/:pathMatch(.*)', component: components.Home, name: 'catch-all' },
+  {
+    path: '/catch-all-redirect/:pathMatch(.*)*',
+    redirect: { name: 'Foo' },
+  },
 ]
 
 async function newRouter(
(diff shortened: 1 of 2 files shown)


mvtandas:
Friendly ping — this fixes the false warning when catch-all routes redirect to named routes. Includes test. Let me know if any changes are needed!

posva:
Thanks but this adds runtime code. Let's go with #2662

PR #2697, opened 2026-05-04
docs: update 'Named Views' examples

skirtles-code:
The CodeSandbox examples aren't currently loading correctly, at least for some people. This may be an intermittent issue, but I think it's worth updating them anyway as they're using some very old dependencies.

I've only updated the two examples in `named-views.md`. There are other CodeSandbox examples elsewhere.

I've migrated the examples to the Vue Playground. This removes many of the external dependencies and makes it easier for other contributors to update the examples.

As CodeSandbox is currently failing I wasn't able to test the old examples properly, though I was able to access the code and used it as the starting point for the new examples.

For the first example, the CodeSandbox was pretty different to the example shown in the docs. I've tried to bring them into closer alignment, while also trying to make the example itself a little more realistic. I've also removed some 'noise' from the docs, such as the `history` option and `class` attributes. Those aren't directly relevant to what's being explained.

For the second example I stayed pretty close to the original CodeSandbox example. I'm mostly happy with it, though I did need to use this workaround to get the initial navigation to work:

```js
router.push('/settings/emails').then(() => {
  createApp(App)
    .use(router)
    .mount('#app')
})
```

## Summary by CodeRabbit

* **Documentation**
  * Updated Named Views guide with revised example outlet naming conventions.
  * Refreshed playground demo links in Named Views and Nested Named Views sections.

### DIFF
--- a/packages/docs/guide/essentials/named-views.md
+++ b/packages/docs/guide/essentials/named-views.md
@@ -8,9 +8,9 @@
 Sometimes you need to display multiple views at the same time instead of nesting them, e.g. creating a layout with a `sidebar` view and a `main` view. This is where named views come in handy. Instead of having one single outlet in your view, you can have multiple and give each of them a name. A `router-view` without a name will be given `default` as its name.
 
 ```vue-html
-<router-view class="view left-sidebar" name="LeftSidebar" />
-<router-view class="view main-content" />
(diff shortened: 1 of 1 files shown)


posva (approved):
You rock! Thanks a ton!

PR #2701, opened 2026-05-06
docs: update 'Nested Routes' example

skirtles-code:
Similar to #2697, I've migrated the example on the **Nested Routes** page from CodeSandbox to the Vue Playground.

The code for the new example is pretty similar to the old code and should match what's described in the docs that precede the link.

## Summary by CodeRabbit

* **Documentation**
  * Updated the demo link in the Nested Routes guide to use Vue Playground instead of CodeSandbox.


posva (approved):
Thanks!

PR #2723, opened 2026-05-29
docs(exp): add modfiy vite config for use custom Parser

WuMingDao:
## not enable

image

image

## enable
image

## my codebase
https://github.com/WuMingDao/vue-router-exp

## Summary by CodeRabbit

* **Documentation**
  * Added a "Setup" section explaining how to enable experimental.paramParsers in the Vite plugin and an updated configuration example.
  * Clarified that src/params is the default parser directory but can be customized.
  * Updated examples to reference the configured param parser directory instead of a hardcoded path.


posva (changes requested):
Thanks, let's create a Setup section after "The problem" that explains how to install

PR #2727, opened 2026-06-01
fix(data-loaders): ignore leaked loader context in components

### LINKED ISSUE #2684: Nested loader is called twice and has itself as a parent when parent loader resolves instantly
### Reproduction

https://stackblitz.com/edit/vitejs-vite-avh6zm6f?file=src%2Fcomponents%2FChildPage.vue,src%2Floaders%2FhomeLoader.ts

1. Open reproduction link
2. Open console, each loader has a `console.log` statement at the top to see when each loader is executed.
3. Verify that on initial load "parent-page" loader is called twice
4. See that first loader is called twice
5. Add breakpoint at `parentLoader`
6. Uncomment line in `ChildPage.vue`.
7. Refresh the preview
8. See that parentLoader is called in infinite loop

### Expected behavior

Each loader is only called once per navigation

### Actual behavior

First loader is called twice. If line in child component is uncommented, first loader is called in infinite loop

### Additional information

Setup includes a helper component that resolves a loader and provides slot with loaded data. 
In my app, I have more than one child page, and so I've added simple caching to avoid refetching parent loader when navigating between them. This simple cache also persists when going back. If you remove caching and force it to reload every time, it will work as intended

posva:
> WIP / draft — opening early for visibility, feedback welcome.

Closes #2684

### Problem

When **all loaders on a route are lazy**, navigation completes while the loaders are still pending. A nested lazy loader (e.g. `child` that `await`s `parent` and then keeps loading) leaves the module-level loader *current context* pointing at its own entry. If a component renders during that window, `useDataLoader()` called in `setup()` reads that stale context as its **parent**, which:

- re-runs the parent loader (the "parent-page" loader is called twice on initial load),
- makes a loader nest under itself (`"…" has itself as parent`),
- and, with two loaders in the same component, spins into an infinite render → reload loop.

The context global is set around a loader's execution and restored on `.finally`; lazy loaders restore it *after* navigation, racing with component rendering.

### Fix

In `useDataLoader`, when invoked from inside a component (`getCurrentInstance()`) and **not** synchronously within a loader function, drop the ambient context — in a component, loaders are always used at the root. A small depth counter (`runInsideLoaderFn` / `isInsideLoaderFn` in `utils.ts`) wraps the actual loader-function call so genuinely nested loaders still nest correctly, even when first triggered synchronously from a component. Applied to both `defineBasicLoader` and `defineColadaLoader`.

### Test

Added a regression test to the shared `tester.ts` (runs for basic + Colada): a parent loader that resolves instantly, a child loader that nests it and keeps loading, and parent/child components that use the loaders directly. Written before the fix and confirmed failing first (`"has itself as parent"` warning + parent loader called twice); passes after.

### Checks

- [x] `pnpm test:unit` — full suite passes (1629)
- [x] `pnpm test:types`
- [x] lint + format clean
- [ ] changelog / docs (if needed)

### DIFF
--- a/packages/router/src/experimental/data-loaders/defineColadaLoader.ts
+++ b/packages/router/src/experimental/data-loaders/defineColadaLoader.ts
@@ -19,11 +19,13 @@ import {
   NavigationResult,
   assign,
   getCurrentContext,
+  isInsideLoaderFn,
   isSubsetOf,
+  runInsideLoaderFn,
   setCurrentContext,
   trackRoute,
 } from './entries/index'
-import { type ShallowRef, shallowRef, watch } from 'vue'
+import { type ShallowRef, getCurrentInstance, shallowRef, watch } from 'vue'
 import {
   type EntryKey,
   type UseQueryOptions,
@@ -133,10 +135,14 @@ export function defineColadaLoader<Data>(
         // needed for automatic refetching and nested loaders
         // https://github.com/posva/unplugin-vue-router/issues/583
         return router[APP_KEY].runWithContext(() =>
-          loader(trackedRoute, {
-            // TODO: provide the query signal too
-            signal: route.meta[ABORT_CONTROLLER_KEY]?.signal,
-          })
+          // mark that we are synchronously inside a loader function so nested
+          // loaders can tell apart their parent context from a leaked one
+          runInsideLoaderFn(() =>
+            loader(trackedRoute, {
+              // TODO: provide the query signal too
+              signal: route.meta[ABORT_CONTROLLER_KEY]?.signal,
+            })
+          )
         )
       },
       key: () => toValueWithParameters(options.key, entry.route.value),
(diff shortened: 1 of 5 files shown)


Truevoly:
Thanks for looking into this issue!

I have prepared another reproduction, this time with more loaders
https://stackblitz.com/edit/vitejs-vite-mqc5mpvr?file=src%2Frouter.js
Every page has two independent loaders, and a nested view with a loader that depends on those two loaders.

If you open console and check logs, you can see that sometimes context is lost, as evident by messages about using `inject` outside of components

I also propose my solution to this, you can find PR [here](https://github.com/vuejs/router/pull/2757)
I would love to hear your feedback and maybe you can spot any mistakes I've made or if my approach can break other things. I have reused your test case from this PR and created another one, slightly modified, to test that reload works as well. My PR is not final solution, as it lacks updates to Colada loader and better tests, but at this moment I want to hear your feedback if you consider my approach reasonable. If yes, I will then bring it up to proper standards

posva:
@Truevoly thanks, your approach broke other tests so I copied over the test to here

PR #2730, opened 2026-06-05
fix(scroll): ignore stale async scrollBehavior results

baozjj:
## Summary
- Track the latest `handleScroll` invocation with `latestScrollId` so only the most recent async `scrollBehavior` result can call `scrollToPosition`
- Apply the same guard in both `createRouter` and `experimental_createRouter`
- Add a regression test covering rapid A → B → C navigations with delayed `scrollBehavior`
## Problem
When `scrollBehavior` returns a Promise (e.g. waiting for DOM or transitions), a superseded navigation could still apply its scroll position to the current page after the user had already navigated elsewhere.
## Test plan
- [x] `pnpm exec vitest run __tests__/scrollBehaviorRace.spec.ts`
- [x] `pnpm test:unit` in `packages/router`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale scroll actions and stale error reports from applying when newer navigations complete first, improving scroll correctness during rapid successive navigations.

* **Tests**
  * Added tests verifying asynchronous scroll behavior and ensuring only the final navigation’s scroll is applied and superseded errors are ignored.


posva (approved):
Nice catch, thanks!

PR #2744, opened 2026-06-30
fix(unplugin): reject invalid hex digits in [x+hh] character codes

mahirhir:
The `[x+hh]` segment encoding checks that the hex buffer is 2 characters long, then decodes it with `parseInt(buffer, 16)`. The problem is `parseInt` stops at the first character that is not a valid hex digit instead of failing, so a buffer where only the second character is invalid is accepted.

For example, `[x+1g]` parses as `0x1` (the `g` is ignored) and silently produces a `U+0001` control character. Meanwhile `[x+ZZ]` and `[x+g1]` are correctly rejected, because there `parseInt` returns `NaN` when the first character is not hex. So the current check only catches cases where the first digit is bad.

This replaces the `NaN`/range check with a `^[0-9a-f]{2}$` test so both characters must be hex digits. The length is already guaranteed to be 2 at that point, and two hex digits are always within `0x00..0xFF`, so the explicit range check is no longer needed. The error message is unchanged.

Added a test for `[x+1g]` alongside the existing `[x+ZZ]` case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for hex-encoded route segments so invalid hex values now fail with a clearer error message.
  * Prevents malformed hex codes from being accepted during parsing.
* **Tests**
  * Added coverage to verify invalid hex input is rejected as expected (including stricter handling for non-hex characters).

### DIFF
--- a/packages/router/src/unplugin/core/tree.spec.ts
+++ b/packages/router/src/unplugin/core/tree.spec.ts
@@ -225,6 +225,14 @@ describe('Tree', () => {
       )
     })
 
+    it('throws when only the second digit is not hex', () => {
+      const tree = new PrefixTree(RESOLVED_OPTIONS)
+
+      expect(() => tree.insert('[x+1g]', '[x+1g].vue')).toThrow(
+        /Invalid hex code "1g"/
+      )
+    })
+
     it('throws on too many digits in hex code', () => {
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #2747, opened 2026-07-13
refactor: remove `@babel/generator`

Norbiros:
> In the spirit of [e18e](https://e18e.dev/), I looked into several popular libraries to identify potential optimization opportunities (this time related to nuxt v5).

`vue-router` depends on `@babel/generator`, despite using it in only one place. Given the package’s download volume, this dependency accounts for more than 1 TB of [package data transferred each month](https://dependents.dev/?package=%40babel%2Fgenerator):

Monthly download statistics for @babel/generator

The single use case in this project can be safely replaced with a small manual implementation, allowing us to remove `@babel/generator` from the production dependency tree. To make sure that change is safe, I added some additional tests related to that scope.

To be clear, this PR was AI assisted, but I manually reviewed and verified the code and logic.

Note: I removed the manual version override for `@babel/generator` introduced in https://github.com/vuejs/router/commit/8d3e60e7ac9fe52f7e664a7ed21bee322517ce13, since `@babel/generator` is now only a development dependency. It also removes double `babel` dependencies between `vue-router` and `vue` itself. I can restore the override if you would prefer to keep it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of arrow-function defaults in route query parameters.
  * Preserved the original function formatting and TypeScript syntax when extracting defaults, including multiline and commented functions.
  * Added support for synchronous, asynchronous, typed, and block-body arrow functions in Vue and TypeScript route definitions.
* **Tests**
  * Expanded coverage to verify serialized defaults and exact source preservation across supported formats.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
   "packageManager": "pnpm@10.30.0",
   "pnpm": {
     "overrides": {
-      "@babel/generator": "^8.0.0-rc.4",
       "@babel/types": "^8.0.0-rc.4",
       "@types/node": "^24.10.4",
       "@vue/compiler-sfc": "^3.5.34",
--- a/packages/router/package.json
+++ b/packages/router/package.json
@@ -136,7 +136,6 @@
     "test:e2e": "playwright test"
   },
   "dependencies": {
-    "@babel/generator": "^8.0.0",
     "@vue-macros/common": "^3.1.3",
     "@vue/devtools-api": "^8.1.5",
     "ast-walker-scope": "^0.9.0",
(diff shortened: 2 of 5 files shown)


posva:
Interesting! I would be very happy to remove the dependency on babel. I'm really surprised I only use it once 😆 I guess it's just because that's the generator.
There is some code manipulation happening in a few places. I wonder if replacing it with something else is possible at some point

posva (approved):
Thanks! I will fix some stuff locally

PR #2751, opened 2026-07-16
chore: move all pnpm settings to `pnpm-workspace.yaml`

btea:
To ease management and avoid conflicts with npm, pnpm supports moving all configurations to the pnpm-workspace.yaml file. For more [github.com/orgs/pnpm/discussions/9037](https://github.com/orgs/pnpm/discussions/9037)

The default value for strictPeerDependencies is false, so no further configuration is required.
https://pnpm.io/10.x/settings#strictpeerdependencies

https://pnpm.io/settings#allowbuilds

## Chores
- Centralized package manager configuration at the workspace level, including forced versions for key dependencies.
- Updated peer dependency handling rules to better support selected tooling builds.
- Simplified the root package and reduced `.npmrc` to only retain ChromeDriver version detection.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,1 @@
-shamefully-hoist=true
 detect_chromedriver_version=true
-strict-peer-dependencies=false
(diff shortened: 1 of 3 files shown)


posva (approved):
Thanks!

PR #2756, opened 2026-07-18
fix: preserve encoded hashes when re-resolving routes

Mateleo:
Related to https://github.com/nuxt/nuxt/issues/32774, as I wanted to fix this issue.

During startup, Nuxt can resolve the initial URL and then pass the normalized route back to `router.replace()`:

```ts
router.replace({ ...resolvedInitialRoute, force: true })
```
At that point, the route's `hash` is decoded while `fullPath` still contains its original encoding. Re-resolving the route rebuilt `fullPath` from the decoded hash, causing encoded characters such as `%26` to become `&` in the browser URL.

The same encoding loss could happen when calling `router.replace()` with a string.

## Changes

Reuse the encoded hash from `fullPath` when it decodes to the current route hash. If the hash was explicitly changed, it continues to be encoded normally.

This keeps encoded fragments intact when replacing or re-resolving normalized routes.

## Tests
I tried to add some coverage, but this is the part where i'm not so sure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL hash handling in routing.
  * Encoded hash values are now correctly decoded for the active route while preserving their encoded form in the full URL.
  * Preserves hash formatting consistently when re-resolving routes or forcing navigation replacements.
* **Tests**
  * Added coverage for hash encoding/decoding behavior during `router.replace()` scenarios involving `%26` and forced updates.

### DIFF
--- a/packages/router/__tests__/router.spec.ts
+++ b/packages/router/__tests__/router.spec.ts
@@ -400,6 +400,31 @@ describe('Router', () => {
     expect(router.currentRoute.value.fullPath).toBe('/#%2526')
   })
 
+  it('keeps an encoded hash when replacing with a string', async () => {
+    const { router } = await newRouter()
+    await router.replace('/#a=1%26b=2')
+    expect(router.currentRoute.value.hash).toBe('#a=1&b=2')
+    expect(router.currentRoute.value.fullPath).toBe('/#a=1%26b=2')
+  })
+
+  it('keeps an encoded hash when re-resolving a resolved route', async () => {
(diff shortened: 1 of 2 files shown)


posva:
Thanks but this behavior is initially intended. With the introduction of params from the hash, I think I can align the behavior of route.hash with URL and leave both `%26` and `&` as is while still decoding in params. I opened a draft at https://github.com/vuejs/router/pull/2760 but this will be a breaking change

PR #2757, opened 2026-07-20
fix(view): implement better context management for loaders

### LINKED ISSUE #2684: Nested loader is called twice and has itself as a parent when parent loader resolves instantly

Truevoly:
close #2684

This PR is in draft, feel free to provide comments, suggestions, critique 

This is my approach to resolve the issue with nested loader's context
I have found another way to reproduce the issue with context and that is by doing reload from component
Since provided solution by posva only updated `useDataLoader` and not `load` function (which is used when doing reload), I was worried that fix will not work for `reload`

[Take a look at this stackblitz](https://stackblitz.com/edit/vitejs-vite-mqc5mpvr?file=src%2FApp.vue). 
There are 4 pages, each page has two loaders that do not depend on each other, and a subpage, that has a loader that depends on those two. Difference between pages is how long it takes for each loader to resolve with a value. 
Open console and navigate between pages, click reload after reloading has finished. While page will not go into infinite loop, it still logs a warning about incorrect context.

[This stackblitz contains vue-router built from this branch](https://stackblitz.com/edit/vitejs-vite-ebt2ibax?file=package.json,.gitignore). Open console and check logs, navigate between pages, reload, try and break it (and I mean it, I want this to be bulletproof)

I've started investigating and trying to follow the context during the different phases of execution. As it turns out, since context is only relevant for nested loaders, we can keep it empty for the most time, and only update on demand.

If our loader has nested loaders, it will call `useDataLoader`. When nested loader is finished, we reach `finally`. Here, we don't know if there are other nested loaders after us, so in case there are, we prepare context for them. We do it only if we're in a nested loader (`parentEntry` is not undefined). Now, context is restored back to original loader's context.
After `finally` finishes, promise is resolved, and two jobs are put into queue original loader and restoration of context back to empty. After doing some sync work, original loader can potentially call other nested loader, which will receive proper context. After that task is done, context is resolved. Any other call will receive empty context

During testing I've spotted a problem with this solution: if you have two nested loaders and second loader resolves before first, when dependent loader reaches second loader it triggers it again. So I've changed the condition on when the loader should be invoked. But now, when doing reload, nested loaders are never triggered. So I force it by resetting their `to` property, so the condition matches and loader is reloaded. I myself am not a fan of this solution, I welcome any suggestion on how to make it better
 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nested data-loader behavior to prevent context leakage during concurrent loading.
  * Fixed reload handling so child loaders properly refetch their data.
  * Improved recovery and context restoration when loader requests fail.

* **Tests**
  * Added coverage for nested lazy-loader loading and reload scenarios, including warning prevention and expected loader execution counts.

### DIFF
--- a/packages/router/src/experimental/data-loaders/defineLoader.ts
+++ b/packages/router/src/experimental/data-loaders/defineLoader.ts
@@ -424,7 +424,7 @@ export function defineBasicLoader<Data>(
         }
       })
 
-    setCurrentContext(currentContext)
+    setCurrentContext([])
     return Object.assign(promise, useDataLoaderResult)
   }
 
--- a/packages/router/src/experimental/data-loaders/defineLoader.ts
+++ b/packages/router/src/experimental/data-loaders/defineLoader.ts
@@ -250,7 +250,10 @@ export function defineBasicLoader<Data>(
         }
       })
       .finally(() => {
-        setCurrentContext(currentContext)
+        // if this is a nested loader, context will be reset in useDataLoader
+        // if this is a root loader, of if it is called from reload or navigation guard,
+        // context is already reset in sync part
+
         // console.log(
         //   `😩 restored context ${options.key}`,
         //   currentContext?.[2]?.fullPath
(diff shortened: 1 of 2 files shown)


posva:
Thanks! See https://github.com/vuejs/router/pull/2727#issuecomment-5056857084

PR #2759, opened 2026-07-20
feat(volar): narrow `typeof useRoute` in a type context

### LINKED ISSUE #2758: `typeof useRoute` isn't narrowed to the file's routes in a type context (only the call is)
In a page SFC using experimental typed routes (with the `sfc-typed-router` Volar plugin), a `useRoute()` call is automatically narrowed to the routes the current file can render. But a manually written `typeof useRoute` (type context) is not. It falls back to the generic default (every route).

```ts
// in src/pages/users/sub-[first]-[second].vue
const route = useRoute()
//    ^? RouteLocationNormalizedLoadedTyped<RouteNamedMap, '/users/sub-[first]-[second]'>   ✅ was already narrowed

type R = ReturnType<typeof useRoute>
//   ^? RouteLocationNormalizedLoadedTyped<RouteNamedMap, keyof RouteNamedMap>              ❌ every route
```

### Reproduction

Impossible on Vue SFC Playground. See linked PR and go to `packages/playground-file-based/src/pages/users/sub-[first]-[second].vue` in IDE. Before rebuilding the `sfc-typed-router` Volar plugin, `RouteFromType` should be any route. After rebuilding it, it should be the narrowed type. `RouteOther` should be ignored.

### Expected behavior

`typeof useRoute` without explicit type arguments should resolve the same as the call: narrowed to the current file's routes.

```ts
type R = ReturnType<typeof useRoute>
//   ^? RouteLocationNormalizedLoadedTyped<RouteNamedMap, '/users/sub-[first]-[second]'>
```

When a generic is passed explicitly (`ReturnType<typeof useRoute<'/b'>>`), we should ignore though.

### Actual behavior

It doesn't narrow.

### Additional information

The Volar plugin already does exactly this for `useRoute()` call expressions, type-query positions just aren't handled yet.

Anoesj:
Closes #2758 (see issue for reproduction steps)

The `sfc-typed-router` Volar plugin already rewrites `useRoute()` **calls** so their return type is narrowed to the routes the current file can render. This extends the same treatment to `typeof useRoute` in a **type context**.

Before this PR, without type arguments, `typeof useRoute` resolves to the function's generic default (`keyof RouteMap`, i.e. every route):

```ts
const route = useRoute()
//    ^? RouteLocationNormalizedLoadedTyped<RouteNamedMap, '/users/sub-[first]-[second]'>   ✅ was already narrowed

type R = ReturnType<typeof useRoute>
//   ^? RouteLocationNormalizedLoadedTyped<RouteNamedMap, keyof RouteNamedMap>              ❌ every route
```

## Change

Handle the `typeof useRoute` type query the same way as the call expression.
Explicit type arguments are untouched, so `ReturnType<typeof useRoute<'/b'>>` still resolves to `RouteLocationNormalizedLoadedTyped<RouteNamedMap, "/b">`.

## Summary by CodeRabbit

- **New Features**
  - Enhanced typed routing in Vue `<script setup>` by improving how `typeof useRoute` and `useRoute()` return types are inferred for file-based routes.
  - Updated the file-based routing playground to show the route name, path, and dynamic `first`/`second` parameters with stronger type checking.

- **Tests**
  - Added automated coverage to confirm typed-router narrowing behavior in TypeScript, preserves explicitly provided types, and skips narrowing for JavaScript.

### DIFF
--- a/packages/playground-file-based/src/pages/users/sub-[first]-[second].vue
+++ b/packages/playground-file-based/src/pages/users/sub-[first]-[second].vue
@@ -1,5 +1,17 @@
-<script lang="ts" setup></script>
+<script lang="ts" setup>
+import { useRoute } from 'vue-router'
+
+const route = useRoute()
+type Route = typeof route
+type RouteFromType = ReturnType<typeof useRoute>
+type RouteOther = ReturnType<typeof useRoute<'/about'>>
+
+route.params.first satisfies RouteFromType['params']['first']
+route.params.second satisfies RouteFromType['params']['second']
+</script>
 
 <template>
   <h1>{{ String($route.name) }} - {{ $route.path }}</h1>
+  <p>{{ $route.params.first satisfies RouteFromType['params']['first'] }}</p>
+  <p>{{ $route.params.second satisfies RouteFromType['params']['second'] }}</p>
 </template>
--- a/packages/router/src/volar/entries/sfc-typed-router.spec.ts
+++ b/packages/router/src/volar/entries/sfc-typed-router.spec.ts
@@ -17,11 +17,12 @@ function transform(
   code: string,
   { lang = 'ts', fileName = FILE_NAME } = {}
 ): string {
-  const instance = plugin({
+  const result = plugin({
     compilerOptions: { rootDir: ROOT_DIR },
     modules: { typescript: ts },
     config: { options: { rootDir: ROOT_DIR } },
   } as any)
+  const instance = Array.isArray(result) ? result[0] : result
 
   const ast = ts.createSourceFile(
     fileName,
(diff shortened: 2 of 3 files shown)


packages/playground-file-based/src/pages/users/sub-[first]-[second].vue
```diff
@@ -1,5 +1,17 @@
-<script lang="ts" setup></script>
+<script lang="ts" setup>
+import { useRoute } from 'vue-router'
+
+const route = useRoute()
+type Route = typeof route
+type RouteFromType = ReturnType<typeof useRoute>
```
Comment (Anoesj) [inline packages/playground-file-based/src/pages/users/sub-[first]-[second].vue:6]: I just picked an arbitrary file to try this in. If you want it moved somewhere else, LMK!

Comment (posva) [reply packages/playground-file-based/src/pages/users/sub-[first]-[second].vue:6]: no worries, it's fine, I use the playground as a dump of tries TBH 😅

posva (approved):
Thanks!

PR #2764, opened 2026-07-27
docs: fix `@see` link to `MatcherPattern` in param parsers

btea:
Use `{@link ...}` so the type reference resolves properly, and drop the `@ts-ignore` that worked around the broken link.

## Summary by CodeRabbit

* **Documentation**
  * Improved API documentation links for parameter parsing.
  * Removed an unnecessary TypeScript suppression comment.


posva (approved):
Thanks!

PR #2774, opened 2026-08-09
perf(link): skip reactivity when rendering on the server

danielroe:
> [!NOTE]
> spotted when doing some benchmarking of `<NuxtLink>` that there was an opportunity to speed up rendering `<RouterLink>` for SSR.

on the server a `<RouterLink>` doesn't need to be reactive, but five computed values are produced by `reactive(useLink(props))`, which has a cost which we can avoid

to avoid two copies of most of the logic, this moves most computed values into plain functions which can be directly executed in the SSR fast-path, or built into computed values for client-side state.

I've done benchmarking locally using 200 links with `renderToString` (note that the mean value here is inflated by GC so the min value is truer unless using with lots of links)

| | min | mean | hz |
| - | - | - | - |
| main | 0.563ms | 0.910ms | 1,100 |
| this pr | 0.384ms | 0.416ms | 2,404 |
| plain `<a>` | 0.147ms | 0.176ms | 5,679 |

my tentative headline is that renders `<RouterLink>` in server environments **~1.5-2.2x faster**. per link that's still tiny: 4.55 μs -> 2.08 μs, but every little helps, right?

(you'll probably want to run your own numbers but I'd be very happy to share my benchmark script if that helps)

let me know if I've missed anything, or you think this is the wrong approach. we can also resolve this in nuxt separately but thought this would be useful to upstream.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server-rendered links to match client-rendered behavior, including navigation, active states, CSS classes, and slot content.
  * Ensured links render consistently whether browser detection is enabled or disabled.
  * Improved support for nested, aliased, and parameterized routes during server-side rendering.

* **Tests**
  * Added comprehensive coverage for server-rendered link scenarios and parity with reactive client rendering.

### DIFF
--- a/packages/router/__tests__/RouterLink.ssr.spec.ts
+++ b/packages/router/__tests__/RouterLink.ssr.spec.ts
@@ -1,0 +1,261 @@
+/**
+ * @vitest-environment node
+ */
+import { createSSRApp, h, type Component } from 'vue'
+import { renderToString } from '@vue/server-renderer'
+import { vi, describe, expect, it } from 'vitest'
+import { START_LOCATION_NORMALIZED } from '../src/location'
+import type { RouteRecordNormalized } from '../src/matcher/types'
+import type {
+  RouteLocationNormalized,
+  RouteLocationResolved,
+} from '../src/typed-routes'
+import type { RouterOptions } from '../src/router'
+
+const records = {
+  home: { path: '/home' } as RouteRecordNormalized,
+  parent: { path: '/parent' } as RouteRecordNormalized,
+  child: { path: '/parent/child' } as RouteRecordNormalized,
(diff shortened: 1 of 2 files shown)


posva:
If this applies to all computed, maybe we could directly apply the improvement to vue. What do you think?

We could also create an SSR RouterLink that only renders an `<a>`, my concern being the final client size increasing with this change

danielroe:
the speedup is largely from avoiding allocating the computed at all (e.g. no arrow function, no proxies), so I'm not sure we can achieve that in Vue itself.

posva:
I see. Let's hold on for the moment on this one: I'm not sure the added size and complexity here is worth it

PR #2784, opened 2026-08-13
build: include nostics in esm

btea:
refer to https://github.com/vuejs/router/commit/94dca111469aeea086c4d7b7767d2186c05916d4

Opening the playground link directly in the readme will result in the following error.
`Failed to resolve module specifier "nostics".Tip: edit the "Import Map" tab to specify import paths for dependencies.`


posva:
Thanks for noticing, I updated the playground links

PR #2786, opened 2026-08-16
docs: access RouterLink spread props via props

### LINKED ISSUE #1429: Extending RouterLink example does not compile in TypeScript
### Version
4.0.15

### Reproduction link
[stackblitz.com](https://stackblitz.com/edit/vue3-vue-router-meta-layout-z1a2yv?file=src/components/AppLink.vue)

### Steps to reproduce
There's an issue with reactivity in the template when extending the routerlink as per example from the docs: https://router.vuejs.org/guide/advanced/extending-router-link.html

The culprit is the `...RouterLink.props` prop spread, which (for some reason) makes props unavailable to the template unless you destructure the props with `toRefs`.

I'm not sure if this a bug in vue-router, a bug in vue, or a limitation/side-effect in general, but the example won't properly build and will throw TS errors (it also throws errors in dev, as you can see in the reproduction).

Uncomment the commented out line with `toRefs` to make it all work.

### What is expected?
Props to be available in the template with having to destructure them and use `toRefs`

Due to `...RouterLink.props` the props are no longer directly available in the template.

---
Again: not sure if it's a bug, or a limitation of sorts.
At least it would be good to add this to the docs if it isn't a bug.

rajanpanth:
## Summary

- update the `<script setup>` RouterLink extension example to use `props.to`, `props.activeClass`, and `props.inactiveClass` in the template
- document why TypeScript users should access spread `RouterLink.props` through the `props` object
- mirror the example and note in the Chinese docs page

Fixes #1429.

## Validation

- `pnpm install --frozen-lockfile --store-dir D:\grand-final prr\.cache\pnpm-store-vue-router`
- `pnpm exec oxfmt --check packages/docs/guide/advanced/extending-router-link.md packages/docs/zh/guide/advanced/extending-router-link.md`
- `pnpm --filter ./packages/docs run docs:translation:compare zh`
- `git diff --check`

## Notes

`pnpm run docs:build` was also attempted. It failed on existing unrelated Twoslash/module-resolution errors in `packages/docs/data-loaders/*` pages (`Cannot find module 'vue-router/experimental'`), after the edited RouterLink docs were processed.

## Summary by CodeRabbit

* **Documentation**
  * Updated the English and Chinese router-link extension examples to use the correct `props` access pattern in Composition API templates.
  * Clarified TypeScript template-binding behavior when working with spread router-link properties.
  * Improved examples for external-link targets and active/inactive class bindings.

### DIFF
--- a/packages/docs/guide/advanced/extending-router-link.md
+++ b/packages/docs/guide/advanced/extending-router-link.md
@@ -32,7 +32,7 @@ const isExternalLink = computed(() => {
 </script>
 
 <template>
-  <a v-if="isExternalLink" v-bind="$attrs" :href="to" target="_blank">
+  <a v-if="isExternalLink" v-bind="$attrs" :href="props.to" target="_blank">
     <slot />
   </a>
   <router-link
@@ -45,13 +45,18 @@ const isExternalLink = computed(() => {
       v-bind="$attrs"
       :href="href"
       @click="navigate"
-      :class="isActive ? activeClass : inactiveClass"
+      :class="isActive ? props.activeClass : props.inactiveClass"
     >
       <slot />
     </a>
   </router-link>
 </template>
 ```
+
+When using `<script setup>` with TypeScript, access props spread from
+`RouterLink.props` through the `props` object in the template, as shown above.
+This avoids relying on the compiler exposing those spread props as individual
+template bindings.
 
 ```vue [Options API]
 <script>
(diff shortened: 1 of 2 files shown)


posva:
Thanks for the try but this still fails as there are other things missing to make TS work well. 

> pnpm run docs:build was also attempted. It failed on existing unrelated Twoslash/module-resolution errors in packages/docs/data-loaders/* pages (Cannot find module 'vue-router/experimental'), after the edited RouterLink docs were processed.

You need to build the router packages

PR #2788, opened 2026-08-18
perf: avoid depending on the current route to resolve absolute locations

danielroe:
another perf one! 🔥

spotted this whilst profiling `<NuxtLink>`. every mounted `<RouterLink>` fully re-resolves its `to` on every navigation, regex matching and param decoding included, even when it's not relative to the current path

this is because `resolve` always reads `currentRoute`, even though it won't use it, so the `route` computed value ends up tracking `currentRoute` as a dependency

👉 **this means that every navigation triggers a fresh `resolve` call for every mounted `RouterLink`**

the total time for navigation is very small both before and after, but _technically_ this is something like a ~2x faster navigation if you use non-relative links....

<details><summary><strong>results of a quick `happy-dom` benchmark</strong></summary>
<p>

| links | navigate (before) | navigate (after) | resolve calls per nav |
| - | - | - | - |
| 50 | 0.116ms | 0.051ms | 50 -> 0 |
| 300 | 0.722ms | 0.323ms | 300 -> 0 |
| 1000 | 2.429ms | 1.341ms | 1000 -> 0 |

</p>
</details> 

this PR tweaks the implementation to read `currentRoute` only when the location actually depends on it.

I've also had to add a `routesVersion` counter because at the moment `computed(() => router.resolve('/late'))` goes stale after `addRoute('/late')` and only recovers because the next navigation reruns it. (but because we don't rerun `resolve`, we have to _track_ addRoute.)  you may have a better idea...

## Summary by CodeRabbit

* **Bug Fixes**
  * Route links now update their active and exact-active states correctly as navigation changes.
  * Route resolution now refreshes when routes are added, removed, or cleared.
  * Absolute paths remain stable when the current route changes.
  * Explicitly provided locations are respected during route resolution.

### DIFF
--- a/packages/router/__tests__/router.spec.ts
+++ b/packages/router/__tests__/router.spec.ts
@@ -1111,7 +1111,7 @@ describe('Router', () => {
 
     it('uses an explicitly passed currentLocation', async () => {
       const { router } = await newRouter()
-      const current = router.resolve('/parent/child')
+      const current = await loadRouteLocation(router.resolve('/parent/child'))
       expect(router.resolve('child', current).path).toBe('/parent/child')
       await router.push('/foo')
       expect(router.resolve('child', current).path).toBe('/parent/child')
--- a/packages/router/__tests__/router.spec.ts
+++ b/packages/router/__tests__/router.spec.ts
@@ -2,6 +2,7 @@
  * @vitest-environment happy-dom
  */
 import fakePromise from 'faked-promise'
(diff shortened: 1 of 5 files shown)


posva (approved):
This is great! I think the size increase is worth it and it won't be needed in the next version where the navigation won't be implicitly relative

posva:
I'm adding some extra tests for the experimental router, then I will merge it

posva (approved):

PR #2797, opened 2026-09-02
fix(query): don't corrupt query keys that collide with Object.prototype

contactjawad:
A query key that happens to be an `Object.prototype` member gets mangled:

```js
parseQuery('toString=foo')
// { toString: [ƒ toString, 'foo'] }, expected { toString: 'foo' }
```

`parseQuery` accumulates into a plain `{}` and checks for repeated keys with `key in query`. Since `in` walks the prototype chain, `toString` (and `valueOf`, `hasOwnProperty`, …) look like they already exist on the first occurrence, so the value gets wrapped in an array with the inherited function.

I switched the accumulator to `Object.create(null)` — which is what the TODO in the file and the experimental implementation already point at. Added a test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query parsing for parameter names that overlap with built-in object properties, such as `toString`, `valueOf`, and `hasOwnProperty`.

### DIFF
--- a/packages/router/__tests__/parseQuery.spec.ts
+++ b/packages/router/__tests__/parseQuery.spec.ts
@@ -66,6 +66,18 @@ describe('parseQuery', () => {
     })
   })
 
+  it('does not collide with Object.prototype keys', () => {
+    expect(parseQuery('toString=foo')).toEqual({
+      toString: 'foo',
+    })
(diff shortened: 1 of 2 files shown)


posva:
Thanks but see https://github.com/vuejs/router/issues/2658

PR #2798, opened 2026-09-02
fix(matcher): preserve declared path casing when building star routes

contactjawad:
`MatcherPatternPathStar.build()` lowercases the static part of the path:

```js
new MatcherPatternPathStar('/Team').build({ pathMatch: '/123' })
// '/team/123', expected '/Team/123'
```

The constructor only keeps `path.toLowerCase()`, so the original casing is gone by the time `build()` runs. The sibling `MatcherPatternPathStatic` already handles this by keeping the original `path` for building and a separate lowercased copy for matching (its spec asserts `build()` preserves case). I mirrored that: original for `build()`, a private lowercased field for the case-insensitive `match()`. Added a test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Route matching remains case-insensitive while generated paths now preserve the original capitalization of the configured prefix.

* **Tests**
  * Added coverage confirming that path building retains declared capitalization without changing case-insensitive matching behavior.

### DIFF
--- a/packages/router/src/experimental/route-resolver/matchers/matcher-pattern-path-star.ts
+++ b/packages/router/src/experimental/route-resolver/matchers/matcher-pattern-path-star.ts
@@ -19,12 +19,18 @@ export class MatcherPatternPathStar implements MatcherPatternPath<{
   pathMatch: string
 }> {
   private path: string
+  /**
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #2804, opened 2026-09-19
feat: accept a prebuilt matcher in `createRouter`

danielroe:
in nuxt at the moment, we create a fresh router for every request (+ therefore a fresh _matcher_ too).

in most cases, the route table is identical, but there's a cost to rebuilding the matcher, which takes up a significant proportion of rendering a given route as the route table grows....

| routes | share of request CPU in `createRouterMatcher` | throughput with a shared matcher |
| --- | --- | --- |
| 5 | ~0.5% | roughly unchanged |
| 300 | 10-13% | +35-39% |

this PR adds an optional `matcher` to `RouterOptions`, falling back to the previous behaviour if it's missing...

in nuxt we'd consume this under an opt-in flag that would also disallow `addRoute`/`removeRoute` at the type level.

you may well prefer this to live only on the experimental resolver API, which seemed to have a similar performance benefit, but this seemed like a quick and easy opt-in performance improvement without entering experimental territory.

would love to know your thoughts...

## Summary by CodeRabbit

* **New Features**
  * Routers can now reuse an externally created route matcher.
  * Multiple routers can share route-matching behavior and consistently resolve the same routes.

* **Tests**
  * Added coverage for shared matcher support and consistent route resolution.


posva (commented):
We can add this just for Nuxt as it will be replaced by the `resolver` option like you said.

posva (approved):

